PR #22, opened 2024-03-26
Uvars in repl feature implemented

Foxinio:
As summarized in commit message:
    - added new core type constructor
    - added translation in toCore
    - made it so that it raises error everywhere in core it may appear
    - implemented every necessary function

There were some merge conflicts so i fixed them before this pull request

### DIFF
--- a/src/Lang/Core.mli
+++ b/src/Lang/Core.mli
@@ -4,8 +4,6 @@
 
 (** Core Language. *)
 
-(* Author: Piotr Polesiuk, 2023,2024 *)
-
 (** Type-level constructor of a kind of all types. The main purpose of this
   type is to encode kind-system in OCaml's GADT. Values of this types are
   never used. *)
--- a/src/Lang/CorePriv/Effect.ml
+++ b/src/Lang/CorePriv/Effect.ml
@@ -4,8 +4,6 @@
 
 (** Operations on effects *)
 
-(* Author: Piotr Polesiuk, 2023,2024 *)
-
 open TypeBase
 
 (** Join of two effects. Same as TEffJoin constructor, but removes duplicates.
--- a/src/Lang/CorePriv/SExprPrinter.ml
+++ b/src/Lang/CorePriv/SExprPrinter.ml
@@ -4,8 +4,6 @@
 
 (** Translating Core to S-expressions *)
 
-(* Author: Piotr Polesiuk, 2023,2024 *)
-
 open TypeBase
 open SExpr
 
--- a/src/Lang/CorePriv/Subst.ml
+++ b/src/Lang/CorePriv/Subst.ml
@@ -4,8 +4,6 @@
 
 (** Type substitutions *)
 
-(* Author: Piotr Polesiuk, 2023,2024 *)
-
 open TypeBase
 
 module Sub = TVar.Map.Make(struct type 'k t = 'k typ end)
--- a/src/Lang/CorePriv/Type.ml
+++ b/src/Lang/CorePriv/Type.ml
@@ -4,8 +4,6 @@
 
 (** Operations on types *)
 
-(* Author: Piotr Polesiuk, 2023,2024 *)
-
 open TypeBase
 
 (** Unit type *)
--- a/src/Lang/CorePriv/WellTypedInvariant.ml
+++ b/src/Lang/CorePriv/WellTypedInvariant.ml
@@ -4,8 +4,6 @@
 
 (** Internal type-checker of the Core language *)
 
-(* Author: Piotr Polesiuk, 2023,2024 *)
-
 open Syntax
 open TypeBase
 
--- a/src/ToCore/Type.ml
+++ b/src/ToCore/Type.ml
@@ -4,8 +4,6 @@
 
 (** Translation of types *)
 
-(* Author: Piotr Polesiuk, 2023,2024 *)
-
 open Common
 
 let eff_cons env x eff =
--- a/src/Config.ml
+++ b/src/Config.ml
@@ -4,8 +4,6 @@
 
 (** Global configuration of the interpreter *)
 
-(* 2024: Patrycja Balik: initial implementation *)
-
 let prelude_path =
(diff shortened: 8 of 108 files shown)


src/Lang/Core.mli
```diff
@@ -53,6 +53,9 @@ type _ typ =
   | TUnit : ktype typ
     (** Unit type *)
 
+  | TUVar : UID.t * 'k kind -> 'k typ
+    (** UVar type kept for pretty printing *)
```
ppolesiuk, line 54: Are they really kept for pretty-printing? Or they are just placeholders for unification variables left by type-inference, and makes sense only in REPL. The invariant about this construct should be said explicitly: they might appear only in REPL mode, and once they appear in the program, information about types in Core becomes inaccurate, and all the type-based transformations (including `WellTypedInvariant`) become unavailable.

src/Lang/CorePriv/Effect.ml
```diff
@@ -34,6 +34,10 @@ let prog_effect = TEffJoin(io, nterm)
 (** Check if the effect is pure *)
 let rec is_pure eff =
   match eff with
+  | TUVar _ ->
+    InterpLib.Error.report ~cls:FatalError
```
ppolesiuk, line 38: It is rather an internal error (see `InterpLib.InternalError` module) than a regular error. It should never happen, and once it is raised it is information to developers that an internal invariant were broken. You can use also `failwith` function for raising internal errors.

src/Lang/CorePriv/SExprPrinter.ml
```diff
@@ -28,6 +28,9 @@ let tr_var  x = Sym (Var.unique_name x)
 let tr_tvar (x : _ tvar) =
   Sym (Printf.sprintf "tp%s" (UID.to_string x.uid))
 
+let tr_uvar (x : UID.t) =
+  Sym (Printf.sprintf "weak%s" (UID.to_string x))
```
ppolesiuk, line 32: It should be `weak#42` or `uvar#42`? At the beginning I thought that `uvar#42` would be better, but now I'm not convinced.

src/Lang/CorePriv/Subst.ml
```diff
@@ -36,6 +36,10 @@ let rec in_type_rec : type k. t -> k typ -> k typ =
   fun sub tp ->
   match tp with
   | TUnit | TEffPure -> tp
+  | TUVar _ ->
+    InterpLib.Error.report ~cls:FatalError
```
ppolesiuk, line 40: Again, rather an internal error than a regular one.

src/Lang/CorePriv/Type.ml
```diff
@@ -32,6 +33,12 @@ let subst_type x stp tp =
 let rec equal : type k. k typ -> k typ -> bool =
   fun tp1 tp2 ->
   match tp1, tp2 with
+  | _, TUVar _
+  | TUVar _, _ ->
+    InterpLib.Error.report ~cls:FatalError
```
ppolesiuk, line 40: Same as before. We should ask, what is the philosophy behind raising errors when we encounter unsolved unification variable? Should we always raise an error, or do the best we can (like in e.g. `subeffect` function). I think, for basic functions, like substitution, checking purity, equality, and subtyping we could try to do the best we can: UIDs stored with unification variables makes possible to compare them. But for more complex functions (e.g. `WellTypedInvariant` module) we should definitely raise an internal error.

src/Lang/CorePriv/WellTypedInvariant.ml
```diff
@@ -116,6 +116,10 @@ end
 let rec tr_type : type k. Env.t -> k typ -> k typ =
   fun env tp ->
   match tp with
+  | TUVar _ ->
+    InterpLib.Error.report ~cls:FatalError
```
ppolesiuk, line 120: It should be an internal error

src/ToCore/Type.ml
```diff
@@ -34,11 +34,9 @@ let pack_type (type k) (k : S.kind) (tp : k T.typ) : T.Type.ex =
 let rec tr_effrow_end env (eff : S.Type.effrow_end) =
   match eff with
   | EEClosed -> T.TEffPure
-  | EEUVar _  ->
-    (* TODO: they can be supported, and its especially useful in REPL *)
-    InterpLib.Error.report ~cls:FatalError
-      "Unsolved unification variables left.";
-    raise InterpLib.Error.Fatal_error
+  | EEUVar (_,uvar) ->
+    let uid = Lang.Unif.UVar.uid uvar in
```
ppolesiuk, line 38: In this place we have to options. Either we are in the REPL mode, then we translate the variable as here. Or we are in the batch mode, then we should raise an error.

src/ToCore/Type.ml
```diff
@@ -51,11 +49,9 @@ let rec tr_effrow_end env (eff : S.Type.effrow_end) =
...
   match S.Type.view tp with
   | TUnit    -> T.Type.Ex TUnit
-  | TUVar _  ->
-    (* TODO: they can be supported, and its especially useful in REPL *)
-    InterpLib.Error.report ~cls:FatalError
-      "Unsolved unification variables left.";
-    raise InterpLib.Error.Fatal_error
+  | TUVar (_,uvar)  ->
+    let uid = Lang.Unif.UVar.uid uvar in
+    T.Type.Ex (TUVar (uid, KType))
```
ppolesiuk, line 53: Some comment as above applies here. Moreover, the kind of this variable is not necessarily `KType`. It should be a translated kind of this variable (it could be an unification variable of e.g. arrow kind).

ppolesiuk (changes requested):
I left some comments in the code. Moreover, I have the two following general comments.

1. Now, information about types is inaccurate in the REPL mode, so the internal type-checker (`WellTypedInvariant`) should be disabled in `Pipeline` when we are in the REPL mode.
2. You should update a metadata in comments at the beginning of the modified files (or we should resign from these metadata).

Issue #15, opened 2024-03-23
Simple namespaces


ppolesiuk:
Namespaces are used for grouping definitions. They can be nested, opened, and included. The concrete syntax should be determined. In the future, this mechanism will be combined with dividing a program into separate files (translation units).

PR #29, opened 2024-03-28
Add modules

forell:
Should address #15.

There are a few shortcomings I'm aware of:
* no source code location tracking within the module path data type for more accurate information in errors;
* failure to resolve a path like `X.Y.Z.x` because there is no module `Y` in `X` results in an error message printing the whole path, which is not ideal;
* pretty-printing of named types prefixed by a path is not implemented properly.

### DIFF
--- a/src/Parser/YaccParser.mly
+++ b/src/Parser/YaccParser.mly
@@ -272,9 +272,6 @@ data_rec_rest
 def
 : KW_IMPLICIT TLID implicit_ty_args type_annot_opt
     { make (DImplicit($2, $3, $4)) }
-| KW_LABEL  expr         { make (DLabel $2) }
-| KW_HANDLE expr EQ expr h_clauses      { make (DHandle($2, $4, $5)) }
-| KW_HANDLE expr KW_WITH expr h_clauses { make (DHandleWith($2, $4, $5)) }
 | KW_METHOD expr EQ expr { make (DMethod($2, $4)) }
 | KW_PUB def_10 { make (DPub $2) }
 | def_10 { $1 }
@@ -284,6 +281,9 @@ def_10
 : KW_LET expr EQ expr    { make (DLet($2, $4)) }
 | data_def               { make (DData $1)     }
 | data_rec data_rec_rest { make (DDataRec ($1 :: $2)) }
+| KW_LABEL  expr         { make (DLabel $2) }
+| KW_HANDLE expr EQ expr h_clauses      { make (DHandle($2, $4, $5)) }
+| KW_HANDLE expr KW_WITH expr h_clauses { make (DHandleWith($2, $4, $5)) }
 | KW_MODULE UID def_list KW_END { make (DModule($2, $3)) }
 | KW_OPEN uid_path { make (DOpen $2) }
 ;


src/TypeInference/DataType.mli
```diff
@@ -9,10 +9,11 @@
...
+val open_data : Env.t -> public:bool -> Module.adt_info -> Env.t
 
 (** Check non-recursive definition of ADT *)
-val check_data_def : Env.t -> S.data_def -> Env.t * T.data_def
+val check_data_def : Env.t -> public:bool -> S.data_def -> Env.t * T.data_def
 
 (** Check mutually recursive definitions of ADTs *)
-val check_rec_data_defs : Env.t -> S.data_def list -> Env.t * T.data_def list
+val check_rec_data_defs :
+  Env.t -> public:bool -> S.data_def list -> Env.t * T.data_def list
```
ppolesiuk, line 19: The `public` flag is shared between all mutually recursive types. Should be possible to define public and private types that are mutually recursive? For regular value, such a restriction might be annoying, but not crucial. Mutually recursive values might be private, but some of them might have public aliases. But we do not have type aliases.

BTW, should we have `abstract` visibility flag for datatypes? And should this potential `abstract` flag have something in common with existential types?

src/TypeInference/Pattern.mli
```diff
@@ -16,7 +16,7 @@ open Common
   top-down in a complex patterns, while the environment flows left-to-right.
   *)
 val check_type :
-  env:Env.t -> scope:T.scope -> S.pattern -> T.typ ->
+  env:Env.t -> scope:T.scope -> ?public:bool -> S.pattern -> T.typ ->
```
ppolesiuk, line 19: I wonder if it makes sense to allow `public` flag in patterns, i.e., in `let (pub x, y) = ...` only `x` becomes public.

ppolesiuk (commented):
There are some issues that should be addressed (see comments below). Some of them may require some nontrivial work. On the other hand, features implemented in this PR are crucial for starting work on Fram, so I recommend to merge these changes. You decide.

src/Lang/Surface.ml
```diff
@@ -249,10 +262,19 @@ and def_data =
...
+    (** Definition of a module *)
+
+  | DOpen of module_name path
+    (** Opening a module *)
+
   | DReplExpr of expr
     (** Print type, evaluate, and print the expression, provided by a user in
       REPL. *)
 
+  | DPub of def
```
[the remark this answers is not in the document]
ppolesiuk: I understand that this approach is somehow clean (we didn't have to touch all the existing constructors), but I'm not sure if it would scale to e.g., `abstract` visibility modifier of datatypes, proposed in this review.

src/TypeInference/Pattern.mli
ppolesiuk: Public flag in patterns potentially would solve the problem of inelegant definition of `def` datatype (pub pub problem), and scalability to `abstract` visibility modifier. Definitions like `pub let x = 42` can be considered as a syntactic sugar for `let (pub x) = 42`, while definitions of types would store more complex visibility modifiers (pub, abstract). Or even we can have visibility modifiers to constructors, so `pub data T = A | B` would be desugared to `data pub T = pub A | pub B` while `abstract data T = A | B` would be `data pub T = A | B`.

src/Lang/Surface.ml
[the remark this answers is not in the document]
ppolesiuk: See the comments to `Pattern.mli`.

ppolesiuk:
Another problem, that is not addressed in these changes is how to handle the situation when a public value has a private type. Should we care about that? Or should we raise an error or warning?

src/TypeInference/Pattern.mli
ppolesiuk: Therefore, I opt for getting rid of `DPub` in Surface (in Raw it still makes sense) and desugar it to public flags in patterns. I think it could be done as a separate task.

PR #38, opened 2024-03-30
Adding Operators

MinionJakub:
Changes to lexer, parser, raw  and desugar to add operators early syntax to declare them.


src/Parser/YaccParser.mly
```diff
@@ -4,13 +4,16 @@
...
+   2024: Jakub  Chomiczewski: operators*)
 
 %token<string> LID UID TLID
+%token<string> OP_0 OP_20 OP_30 OP_40 OP_50 OP_60 OP_70 OP_80 OP_90 OP_100
+%token<string> OP_225 
 %token<int> NUM
 %token<string> STR
 %token BR_OPN BR_CLS SBR_OPN SBR_CLS CBR_OPN CBR_CLS
-%token ARROW ARROW2 BAR COLON COMMA DOT EQ SEMICOLON2 SLASH
+%token ARROW ARROW2 BAR COLON COMMA DOT EQ SEMICOLON2 SLASH SEMICOLON
```
ppolesiuk, line 16: Why the `SEMICOLON` is a separate token? Isn't it just a regular operator?

src/Parser/YaccParser.mly
```diff
@@ -153,26 +230,132 @@ expr
...
+: def_list1 KW_IN expr_no_comma  { make (EDefs($1, $3)) }
+| KW_FN expr_250_list1 ARROW2 expr_no_comma { make (EFn($2, $4))   }
+| KW_EFFECT expr_250_list effect_resumption_opt ARROW2 expr_no_comma
+    { make (EEffect($2, $3, $5)) }
+| KW_HANDLER expr_no_comma { make (EHandler $2) }
+| expr_0_no_comma { $1 }
+;
+
+
+//Evaluate first expresion then discard the resulte and after that evaluate second expression
```
ppolesiuk, line 246: What is the role of this comment?

src/Parser/YaccParser.mly
```diff
@@ -153,26 +230,132 @@ expr
...
+| expr_35 { $1 }
+;
+
+expr_30_no_comma
+: expr_30_no_comma COLON ty_expr { make (EAnnot($1, $3)) }
+| expr_35_no_comma { $1 }
+;
+
+expr_35
+// pattern = exp1 , exp2
```
ppolesiuk, line 296: I do not understand the role of `pattern` in this comment. We could remove these comments, or reformulate them to e.g., `expr1 , expr2`. I opt for the second options, as it would help connect operator precedence with precedence of other language constructs.

src/Parser/YaccParser.mly
```diff
@@ -208,17 +395,20 @@ expr_300
...
-| BR_OPN expr BR_CLS { make (EParen $2)    }
-| KW_MATCH expr KW_WITH KW_END { make (EMatch($2, [])) }
+: LID                          { make (EVar $1)       }
+| TLID                         { make (EImplicit $1)  }
+| UNDERSCORE                   { make EWildcard       }
+| NUM                          { make (ENum $1)          }
+| STR                          { make (EStr $1)          }
+| BR_OPN BR_CLS                { make EUnit              }
+| BR_OPN expr BR_CLS           { make (EParen $2)        }
+| KW_MATCH expr KW_WITH KW_END { make (EMatch($2, []))}
```
ppolesiuk, line 405: You changed only whitespace character in this piece of code. We should avoid such a practice, since may cause unnecessary merge conflicts.

src/Parser/YaccParser.mly
```diff
@@ -208,17 +395,20 @@ expr_300
...
 | KW_MATCH expr KW_WITH bar_opt match_clause_list KW_END
   { make (EMatch($2, $5)) }
 | CBR_OPN field_list CBR_CLS { make (ERecord $2) }
+| BR_OPN op BR_CLS             { make (EBOpID ($2).data) }
+| BR_OPN op DOT BR_CLS  { make (EUOpID ($2).data) }
```
ppolesiuk, line 410: Long sequences of spaces that do not align anything.

src/Parser/YaccParser.mly
```diff
@@ -270,20 +460,20 @@ data_rec_rest
 /* ========================================================================= */
 
 def
-: KW_IMPLICIT TLID implicit_ty_args type_annot_opt
+:  KW_IMPLICIT TLID implicit_ty_args type_annot_opt
```
ppolesiuk, line 463: Another unnecessary change. My advice is to run `git diff` before each commit, to find such accidental changes.

src/Parser/YaccParser.mly
```diff
@@ -270,20 +460,20 @@ data_rec_rest
...
     { make (DImplicit($2, $3, $4)) }
-| KW_METHOD expr EQ expr { make (DMethod($2, $4)) }
```
ppolesiuk, line 275: Methods were moved to `def_10`, which means that they may be marked as public now. Is it intentional?

ppolesiuk (approved):
The implementation of new features in this PR is solid, but I have some minor concerns regarding mostly code formatting (and I found a bug in lexer). As there are merge conflicts, and merging this PR into master requires intervention, I can resolve this minor problems myself during a merge process.

src/Parser/YaccParser.mly
MinionJakub: I added this token because I am using this in op_0.

src/Parser/YaccParser.mly
MinionJakub: It was early comment that I should have deleted. It was only for me to think about operators and what they do.

src/Parser/YaccParser.mly
ppolesiuk: But it is completely opposite: it is used in `op_0` because it is a separate token. It can be removed from the list of tokens and it would simplify lexer and `op_0` grammar.

src/Parser/Lexer.mll
```diff
@@ -38,6 +39,44 @@ let kw_map =
   ; "_",        UNDERSCORE
   ] |> List.to_seq |> Hashtbl.of_seq
 
+(* let op_map = Hashtbl.create 32 *)
+let op_map = 
+  let open YaccParser in
+  [ "->", ARROW
+  ; "→",  ARROW
```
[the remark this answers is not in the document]
ppolesiuk: I think we should get rid of unicode arrows instead of moving them to general lexer rules. The reason is that position tracking in ocamllex does not support unicode.

ppolesiuk:
I've manually resolved conflicts, squashed all commits, and merged them into master. The pull request can be closed.

PR #39, opened 2024-04-01
Implemented: Unification variables in REPL

Foxinio:
Implemented: Unification variables in REPL

- changed signatures of tr_program and some other functions to take additional argument whether program is launched as repl or not
- toCore.Env now keeps track of operational mode (repl or not)
- made check_invariant dependent on operational mode
- added new core type constructor
- added translation in toCore
- adapted uvars in core to what i think they should do
- in WellTypedInvariant wherever uvar may be seen it is reported as InternalError
- implemented every necessary function

### DIFF
--- a/src/Lang/Core.mli
+++ b/src/Lang/Core.mli
@@ -48,6 +48,9 @@ end
 
 (** Types, indexed by a type-represented kind *)
 type _ typ =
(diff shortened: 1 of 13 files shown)


ppolesiuk (approved):
Thanks for this solid piece of work. I have nothing to add here.

Issue #31, opened 2024-03-29
`abstract` visibility modifier for datatypes.


ppolesiuk:
Datatype could be marked as abstract, which means that the type variable is publicly visible, but its constructors remains private. The keyword used to mark datatypes as abstract can be discussed, as `abstract` seems to be bit long. Other options that can be considered are `abs` or `abstr`.

Issue #30, opened 2024-03-29
`pub` definition modifier as syntactic sugar for public patterns


ppolesiuk:
As discussed in #29 there should be no `DPub` constructor in Surface. Public definitions should be desugared to definitions with public patterns. This solution seems to be cleaner and scales better to other features that we aim for, including the following.

- Public and abstract visibility modifiers for datatypes.
- Selectively marking as public some of mutually recursive definitions or datatypes.

PR #47, opened 2024-04-09
Clean up visibility implementation and add `abstr`

forell:
Should address issues #30 and #31.

* Addition of "abstract" ADT definitions: `abstr data T = C`. There are still ways to access the constructor, e.g., with an annotation `let f (C : T) = ...`, which should probably be fixed.
* Removal of `DPub` constructor in Surface. Identifiers in `let` and patterns now carry their own visibility, and so do other kinds of definitions and ADT constructors.
* Removeal of `DPub` constructor in Raw. It was no longer useful for (recursive) type definitions, so the implementation is more uniform without it.
* Addition of `pub` in patterns (in Raw only). This means that definitions such as `let Pair (pub x) y = ...` are now legal. It can also appear in places that don't really make sense, such as function parameters: `let f (pub x) = ...`.
* Method definitions can now be preceded by `pub`, but it doesn't do anything yet.

### DIFF
--- a/src/Lang/Surface.ml
+++ b/src/Lang/Surface.ml
@@ -110,10 +110,13 @@ and named_type_arg = (tname * type_arg) node
 
 (** Declaration of constructor of ADT *)
 type ctor_decl = ctor_decl_data node
-and ctor_decl_data =
-  | CtorDecl of is_public * ctor_name *
-                named_type_arg list * named_scheme list * scheme_expr list
-    (** Declaration of constructor of ADT *)
+and ctor_decl_data = {
+  cd_public      : is_public;
+  cd_name        : ctor_name;
+  cd_targs       : named_type_arg list;
+  cd_named       : named_scheme list;
+  cd_arg_schemes : scheme_expr list
+}
 
 (** Definition of ADT *)
 type data_def = data_def_data node
--- a/src/Parser/Desugar.ml
+++ b/src/Parser/Desugar.ml
@@ -209,22 +209,27 @@ let rec tr_type_def (tp : Raw.type_expr) args =
   | TTypeLbl _ | TEffectLbl _ ->
     Error.fatal (Error.desugar_error tp.pos)
 
-let tr_ctor_decl ~public (d : Raw.ctor_decl) =
+let tr_ctor_decl ~public:cd_public (d : Raw.ctor_decl) =
   let make data = { d with data = data } in
   match d.data with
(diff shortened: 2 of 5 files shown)


src/Lang/Surface.ml
```diff
@@ -108,14 +111,14 @@ and named_type_arg = (tname * type_arg) node
 (** Declaration of constructor of ADT *)
 type ctor_decl = ctor_decl_data node
 and ctor_decl_data =
-  | CtorDecl of
-    ctor_name * named_type_arg list * named_scheme list * scheme_expr list
+  | CtorDecl of is_public * ctor_name *
+                named_type_arg list * named_scheme list * scheme_expr list
```
ppolesiuk, line 115: With 5 parameters to the constructor we could consider using inline records.

src/Parser/Desugar.ml
```diff
@@ -209,24 +209,26 @@ let rec tr_type_def (tp : Raw.type_expr) args =
...
-    make (CtorDecl(name, [], [], List.map tr_scheme_expr schs))
+    make (CtorDecl(public, name, [], [], List.map tr_scheme_expr schs))
 
 let tr_data_def (dd : Raw.data_def) =
   let make data = { dd with data = data } in
   match dd.data with
-  | DD_Data(tp, cs) ->
+  | DD_Data(vis, tp, cs) ->
+    let pub_type  = vis <> DV_Private in
+    let pub_ctors = vis =  DV_Public  in
```
ppolesiuk, line 227: I'm not sure how this code is resistant to changes. What would happen, if we add new constructors to `data_vis` type? Will we find this place and change it accordingly, or just overlook it and introduce a bug to code?

src/Parser/Desugar.ml
ppolesiuk, line 1: Implicit parameters would really help here.

src/TypeInference/Env.ml
```diff
@@ -153,7 +153,8 @@ let lookup_method_map env owner =
   | Some map -> map
   | None     -> StrMap.empty
 
-let add_poly_method env owner name sch =
+let add_poly_method ?(public=false) env owner name sch =
+  (* TODO: implement public methods *)
```
ppolesiuk, line 157: Could you add this as an issue?

ppolesiuk (approved):
I approve the changes. However, I have two minor remarks that could be addressed. Could you add the two issues raised by this changes, those mentioned in the description of PR and the TODO comment?

PR #52, opened 2024-04-10
Reporting of escaping type variables

### LINKED ISSUE #16: Error messages about escaping type variables
Escaping type variables are detected, but error messages are not very informative. For instance, the following program
```
let foo f =
  data T = A in
  f A
```
produces the following error.
```
test.dbl:3:5: error: This expression has type T, but an expression was expected of type T
```
We could produce better error messages.

kwasielewski:
This PR uses previously defined `Error.type_escapes_its_scope` function to report escaping type variable errors in Unification module. After clean-up most of the work simplified to propagating position in the source code in order to use mentioned function.  

Fixes #16


ppolesiuk (changes requested):
I'm not sure if the error messages are better than before. For instance, the old message in `test/err/tc_0001_escapingType.dbl` says that the expression has type `#T` which is important for the user (but, it's still not clear, why it cannot be unified with `a`). The new message says only that `#T` escapes its scope, but its not clear why. Is it possible to have both error messages, i.e, that expression has type `#T` and cannot be unified `a` because it would cause `#T` to escape its scope?

I suppose that it would require some changes in the interface of `Unification`. In case of function like `subtype` or `subeffect` we could add additional functional parameter `on_error` which will report main error (then these function could return unit). This `on_error` function could take as a parameter optional notes, that would passed around and finally reach the third component of the tuple `Error.t` For functions like `to_arrow` or `from_handler` is it possible to get escaping variable? If so, we could add additional information in `Arr_No`, `H_No`, and `L_No` constructors.

kwasielewski:
I tried two options:
- first is to simply change `Error.fatal` to `Error.report` to print the information about escaping type variable. In the examples the resulting message may look like 
```
./test/err/tc_0001_escapingType.dbl:3:18: error: Type variable #T escapes its scope
./test/err/tc_0001_escapingType.dbl:3:18: error: This expression has type #T, but an expression was expected of type a
  Type #T is defined at ./test/err/tc_0001_escapingType.dbl:3:10-11
```
I don't like the order of the messages, but I guess it the simplest way to display all needed information.
- second option I considered is to use return extra information about escaping variables (and possibly other things) form functions like `subtype` and then add that information to the final error raised. It's more complicated, so I'll have to double check my solution (especially whether all cases are covered/necessary). Errors in this approach look like this
```
./test/err/tc_0001_escapingType.dbl:3:18: error: This expression has type #T, but an expression was expected of type a
  Type #T is defined at ./test/err/tc_0001_escapingType.dbl:3:10-11
./test/err/tc_0001_escapingType.dbl:3:18: note: Type variable #T escapes its scope
```

src/TypeInference/Unification.mli
```diff
@@ -51,23 +51,23 @@ val kind_to_arrow : T.kind -> (T.kind * T.kind) option
 
 (** Check if two types (of the same kind) are equivalent.
   It performs some unifications when necessary. *)
-val unify_type : Env.t -> T.typ -> T.typ -> bool
+val unify_type : pos:Position.t -> Env.t -> T.typ -> T.typ -> bool
```
ppolesiuk, line 54: The `unify_type` function may also cause a type variable to escape.

src/TypeInference/Unification.ml
```diff
@@ -299,40 +300,57 @@ and check_subscheme env sch1 sch2 =
...
+  match BRef.bracket (fun () -> check_subtype ~pos env tp1 tp2) with
+  | ()              -> true, []
+  | exception Error -> false, []
+  | exception Escapes_scope (p, e, t) ->
+    let pp_ctx = Pretty.empty_context () in
+    let msg = Printf.sprintf
+    "Type variable %s escapes its scope"
+    (Pretty.tvar_to_string pp_ctx env t) in
+    false, [(p, msg)]
+let subscheme ~pos env sch1 sch2 =
```
ppolesiuk, line 336: Empty line between `subtype` and `subscheme` is missing.

ppolesiuk (changes requested):
I like the idea of printing a regular type/effect mismatch error together with a note on escaping variables. However, I don't like the new signature of `subeffect`, `subtype`, and `subscheme` function because of the following reasons.

1. The returned type `bool * (Position.t * string) list` is inelegant and contains redundant values: if the first component is `true` then the second one is always an empty list. Option type provides more precise information. However, using option type might be misleading, as the `None` means 'correct', and `Some` means error. Of course, you can define own type isomorphic to `option` with more meaningful names of constructors.
2. The code is polluted with passing location, that should finally reach `set_uvar` function. This problem is especially exhibited in 'to/from_handler' and 'to/from_label' functions: they need location, but they should never use it. Locations have nothing to do with unification, and within each call of a function from `Unification` module, the location is constant, so why do we need them here?
3. Functions `subeffect`, `subtype`, and `subscheme` are responsible for producing the *text* of an error message. So far, only `Pretty` and `Error` modules were responsible for producing human readable texts. These changes lead to spreading this functionality across more modules, which is obviously bad.

The solution that I could propose is to change the return type of `subeffect`, `subtype`, `subscheme`, and `unify_type` (yes, you forgot to attach the note also there) functions to some ADT (say, named `unify_result`) with two constructors: one without parameters, that indicates that everything is ok, and one for errors. Errors can contain some optional information about escaping variables, but in a more abstract representation, e.g., environment + escaping type variable (but without location, in order to not pollute the code). Module `Error` can contain a new function that would handle such a result. Its signature could be the following.
```
val check_unify_result : pos:Position.t -> unify_error -> on_error:(pos.Position.t -> t) -> unit
```
We can consider adding another parameter that indicates the method of error reporting (`report`/`fatal`).

src/TypeInference/Error.ml
```diff
@@ -286,13 +289,32 @@ let ctor_not_in_type ~pos ~env name tp =
     (Pretty.type_to_string pp_ctx env tp)
   in (pos, msg ^ Pretty.additional_info pp_ctx, [])
 
+let escaping_tvar_message ~env x =
+  let pp_ctx = Pretty.empty_context () in
+    let msg = Printf.sprintf
```
ppolesiuk, line 294: Code formatting could be improved. Let-definitions should be on the same indentation level.

src/TypeInference/Error.mli
```diff
@@ -18,6 +18,9 @@ val report : t -> unit
 (** Report a warning *)
 val warn : t -> unit
 
+(** Add notes to an error *)
+val add_notes : t -> (Position.t * string) list -> t
```
ppolesiuk, line 22: Should this function be public?

src/TypeInference/Error.mli
```diff
@@ -86,7 +89,10 @@ val empty_match_on_nonempty_adt : pos:Position.t -> env:Env.t -> T.typ -> t
 
 val ctor_not_in_type : pos:Position.t -> env:Env.t -> S.ctor_name -> T.typ -> t
 
+val unification_error_to_string : Unification.unification_error -> string
+val escaping_tvar_message : env:Env.t -> T.tvar -> string
 val type_escapes_its_scope : pos:Position.t -> env:Env.t -> T.tvar -> t
+val check_unify_result : is_fatal:bool -> pos:Position.t -> Unification.unification_result -> on_error:t -> unit
```
ppolesiuk, line 95: This line is too long (more than 80 columns).

src/TypeInference/Error.mli
```diff
@@ -86,7 +89,10 @@ val empty_match_on_nonempty_adt : pos:Position.t -> env:Env.t -> T.typ -> t
 
 val ctor_not_in_type : pos:Position.t -> env:Env.t -> S.ctor_name -> T.typ -> t
 
+val unification_error_to_string : Unification.unification_error -> string
+val escaping_tvar_message : env:Env.t -> T.tvar -> string
```
ppolesiuk, line 93: These two function should be private.

src/TypeInference/Expr.ml
```diff
@@ -151,19 +151,20 @@ and method_call_ctx ~pos ~env ~self ~self_tp ~self_r_eff ~eff =
...
-      Error.report
-        (Error.expr_type_mismatch ~pos:self.pos ~env self_tp self_tp');
-    if not (Unification.subeffect env res_eff eff) then
-      Error.report (Error.method_effect_mismatch ~pos ~env res_eff eff);
+    let unification_result = Unification.subtype env self_tp self_tp'
+      in Error.check_unify_result ~is_fatal:false ~pos unification_result
+        ~on_error:(Error.expr_type_mismatch ~pos:self.pos ~env self_tp self_tp');
+    let unification_result = Unification.subeffect env res_eff eff
+      in Error.check_unify_result ~is_fatal:false ~pos unification_result
+        ~on_error:(Error.method_effect_mismatch ~pos:pos ~env res_eff eff);
```
ppolesiuk, line 159: Code formatting is inconsistent with existing code. This comment also applies to other changes in this file and `Pattern.ml`

src/TypeInference/Unification.ml
```diff
@@ -202,7 +209,7 @@ and unify env tp1 tp2 =
   | TLabel(eff1, rtp1, reff1), TLabel(eff2, rtp2, reff2) ->
     unify_at_kind env eff1 eff2 T.Kind.k_effect;
     unify env rtp1 rtp2;
-    unify_at_kind env reff1 reff2 T.Kind.k_effrow;
+    unify_at_kind  env reff1 reff2 T.Kind.k_effrow;
```
ppolesiuk, line 212: Unnecessary addition of space.

src/TypeInference/Unification.mli
```diff
@@ -41,6 +41,14 @@ type label =
   | L_Label of T.effect * T.typ * T.effrow
     (** Label type *)
 
+type unification_result =
```
ppolesiuk, line 44: This name of type is very long, when used in qualified form: `Unification.unification_result`. Since we never open `Unification` module, I suggest to shorten it to just `result`.

src/TypeInference/Unification.mli
```diff
@@ -41,6 +41,14 @@ type label =
...
+  | Unify_Success
+    (** Unification succeeded *)
+  | Unify_Fail of unification_error list
+    (** Unification failed with list of errors*)
+
+and unification_error = TVar_escaped_scope of Env.t * T.tvar
```
ppolesiuk, line 50: Should these two types be defined as mutually recursive?

ppolesiuk (approved):
I'm happy with the changes. However, I have minor remarks regarding code formatting. I'll fix these issues and merge changes to the upstream.

PR #60, opened 2024-04-14
Syntactic sugar for &&, ||, and ;

### LINKED ISSUE #55: Syntactic sugar for `&&`, `||`, and `;`
These operators need special treatment. Operators `&&` and `||` are lazy and can be desugared to conditional expression. Operator `;` is could be defined in Prelude as
```
let (;) () x = x
```
but in `e1 ; e2` expression `e2` should be at tail position. It is better to desugar `e1 ; e2` as `let () = e1 in e2` or `let _ = (e1 : Unit) in e2`.

Florian3k:
Closes #55


src/Parser/Desugar.ml
```diff
@@ -7,33 +7,39 @@
 open Lang.Surface
 
 (** Translation of the binary operator's name to the regular identifier. *)
-let make_bop_id str = "(" ^ str ^ ")"
+let tr_bop_id (op : string node) =
+  if List.mem op.data ["&&"; "||"; ";"] then
+    Error.fatal (Error.desugar_error op.pos)
+  else "(" ^ op.data ^ ")"
```
Florian3k, line 13: Usage of reserved operators in now prohibited outside of binary expressions.

ppolesiuk: There is a question if `(;) (foo ()) (bar ())` is a valid code, but the decision to allow special operators only is binary expression seems reasonable and I like it. I like the way, how it is handled, i.e., in `tr_bop_id` function which make it works in all places. However, the error messages could be improved (but here, it is not too bad).
```
$ cat test4.dbl
data T = (;)
$ dbl test4.dbl
test4.dbl:1:10-12: fatal error: Syntax error. This construction cannot be used in this context
```

src/Parser/Desugar.ml
```diff
@@ -500,12 +508,22 @@ and tr_expr (e : Raw.expr) =
...
-      }, e2))
+  | EBOp(exp1,op,exp2) ->
+    let with_nowhere data = { pos = Position.nowhere; data = data} in
+    begin match op.data with
+    | "&&" ->
+      tr_expr (make (Raw.EIf(exp1, exp2, with_nowhere (Raw.ECtor("False")))))
+    | "||" ->
+      tr_expr (make (Raw.EIf(exp1, with_nowhere (Raw.ECtor("True")), exp2)))
+    | ";" ->
+      tr_expr (make (Raw.EDefs([make (Raw.DLet(false, make Raw.EWildcard, exp1))], exp2)))
```
ppolesiuk, line 519: This way of desugaring does not enforce `exp1` to have `Unit` type. Because forgetting function parameter is a common programmer error, it should be fixed.
```
$ cat test3.dbl
let _ = 42; True
$ dbl test3.dbl
$
```

Moreover, this line does not fit 80 columns.

src/Parser/Desugar.ml
```diff
@@ -500,12 +508,22 @@ and tr_expr (e : Raw.expr) =
...
-    let e1 = tr_expr exp1 and e2 = tr_expr exp2 in 
-    make (EApp(
-      { pos  = Position.join e1.pos op.pos; 
-        data = EApp(tr_bop_to_expr op, e1)
-      }, e2))
+  | EBOp(exp1,op,exp2) ->
+    let with_nowhere data = { pos = Position.nowhere; data = data} in
+    begin match op.data with
+    | "&&" ->
+      tr_expr (make (Raw.EIf(exp1, exp2, with_nowhere (Raw.ECtor("False")))))
```
ppolesiuk, line 515: While for correct programs it works well, the error messages for wrong programs should be improved. First of all, a type annotation for `exp1` would help.
```
$ cat test1.dbl
let _ = 42 && True
$ dbl test1.dbl
test1.dbl:1:9-18: fatal error: This pattern matches values of type Int, which is not an ADT
```
The other problem is that `False` is located nowhere.
```
$ cat test2.dbl
let _ = True && 42
$ dbl test2.dbl
<nowhere>:1:1: error: This expression has type Bool, but an expression was expected of type Int
```
Adding type annotation to `exp2` should help. Alternatively, we could inline desugaring `EIf` to `EMatch` and revers the order of clauses.

test/ok/ok0093_specialBinops.dbl
ppolesiuk, line 1: I like this test, but some lines are longer that 80 columns.

ppolesiuk (changes requested):
Thanks for solid piece of work. I have minor concerns regarding location tracking and error reporting. Moreover some lines do not fit in 80 columns.

Florian3k:
After improvements:

```
$ dbl parser_0000_illegalBinopPattern.dbl 
parser_0000_illegalBinopPattern.dbl:1:8-9: fatal error: Syntax error. Operator && can only be used in binary expressions

$ dbl parser_0001_illegalBinopCtor.dbl   
parser_0001_illegalBinopCtor.dbl:1:10-12: fatal error: Syntax error. Operator ; can only be used in binary expressions

$ dbl parser_0002_illegalBinopMethod.dbl 
parser_0002_illegalBinopMethod.dbl:1:1-14: fatal error: Syntax error. Operator || can only be used in binary expressions

$ dbl tc_0005_specialBinops.dbl 
tc_0005_specialBinops.dbl:1:9-10: error: This expression has type String, but an expression was expected of type Bool
tc_0005_specialBinops.dbl:2:9-10: error: This expression has type String, but an expression was expected of type Bool
tc_0005_specialBinops.dbl:2:15-16: error: This expression has type String, but an expression was expected of type Bool
tc_0005_specialBinops.dbl:2:21-22: error: This expression has type String, but an expression was expected of type Bool
tc_0005_specialBinops.dbl:3:17: error: This expression has type Int, but an expression was expected of type Bool
tc_0005_specialBinops.dbl:4:9-10: error: This expression has type String, but an expression was expected of type Unit
```

ppolesiuk (approved):
Now, it looks good. However, code formatting could be changed, to make it more consistent with existing code.

PR #61, opened 2024-04-15
Unary conditional expression

### LINKED ISSUE #56: Syntactic sugar for unary conditional expression
Expression `if e1 then e2` could be desugared to `if e1 then e2 else ()`.

Florian3k:
Closes #56

### DIFF
--- a/src/Parser/YaccParser.mly
+++ b/src/Parser/YaccParser.mly
@@ -272,12 +272,36 @@ effect_resumption_opt
 ;
 
 expr_10
-: KW_IF expr KW_THEN expr_10 KW_ELSE expr_10 { make (EIf($2, $4, $6)) }
-| expr_20 { $1 }
+: expr_15_open { $1 }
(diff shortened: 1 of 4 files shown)


src/Parser/YaccParser.mly
```diff
@@ -272,12 +272,36 @@ effect_resumption_opt
...
+| expr_15_closed { $1 }
 ;
 
 expr_10_no_comma
-: KW_IF expr_no_comma KW_THEN expr_10_no_comma KW_ELSE expr_10_no_comma{ make (EIf($2, $4, $6)) }
+: expr_15_no_comma_open { $1 }
+| expr_15_no_comma_closed { $1 }
+;
+
+expr_15_open
```
ppolesiuk, line 284: I suggest to rename these symbols to `expr_10_open`, `expr_10_closed`, etc.. They are at the same level as `expr_10`.

PR #63, opened 2024-04-15
Wildcard in type parameters

wbukowski:
Adds wildcards in ADT parameters.

Addresses #7


ppolesiuk (changes requested):
Technically treating wildcard as type variables of empty name is correct, but I would like to not introduce such code into repository, because of the the following reasons.

1. The name `""` would not collide with anything else, because lexer would not allow to create user-defined variables of such a name. It is a hack.
2. It would not work well, when pretty printer will print such a variable in case of an error. Currently I don't see any possibility to enforce pretty-printer to do that, but if we add explicit type binders in actual function parameters (which have scheme instead of type), it would be different. For instance, if we have a function `foo : ({T}->T->T) -> Unit` and we call it, using wildcard as a name of `T` (currently it is not possible): `foo (fn {X=_} x => 42)` then we would get an error, that uses the name `""`.

I suggest to introduce wildcard in Surface, and use `Env.add_anon_tvar` in the type-checker.

ppolesiuk (approved):
I'm happy with these changes. I approve to merge them to the upstream.

PR #65, opened 2024-04-29
Recursion

### LINKED ISSUE #5: Recursive functions
Recursive functions are not implemented yet (but they are expressible via recursive types: see `fix` in the Prelude). We should add (mutually) recursive functions to all intermediate languages in the pipeline (`Surface`, `Unif`, `Core`, and `Untyped`).

ppolesiuk:
Fixes #5 

Current implementation have some limitations:
- Only mutually recursive ADTs and mutually recursive functions/methods
  are supported
- Recursive methods do not work yet, due to limited guessing of the type
  of recursive value before actual type-checking
- Recursive values of ADTs guarded by a function do not work yet
  (fully-applied constructor is an application, so its not recognized as
  a productive value)
- Type-checker does not check the name uniqueness of mutually recursive
  definitions (except mutually recursive types)
- Internal type-checker of Core doesn't check if recursive functions
  have `NTerm` effect on correct position.

I need to refactor the type-checker before fixing most of these issues. But I hope the current implementation of recursion pushes us forward, so it could be merged into upstream.


src/Utils/Var.ml
```diff
@@ -22,4 +22,6 @@ let fresh ?(name="x") () =
 let unique_name x =
   x.name ^ UID.to_string x.uid
 
+let equal x y = x == y
```
forell, line 25: I don't understand why physical equality is used here.

test/ok/ok0046_mutuallDataRec.dbl
forell, line 1: Btw, there's a typo in this test's name (should be "mutual", I think).

forell (commented):
Aside from my minor comments, I think the changes make sense to be merged in the current state. My only design-related question, about how `pub rec ... end` should interact with visibility modifiers within the block, can be raised at today's project meeting.

Issue #69, opened 2024-04-30
Mixed mutually recursive blocks


forell:
We should be able to mix data type and function definitions in a single `rec ... end` block. Additionally, other kinds of definitions could be allowed in these blocks, such as module definitions.

PR #76, opened 2024-05-03
Refactoring of `TypeInference`

ppolesiuk:
- Split large mutually-recursive block in `TypeInference.Expr` into several modules: `Def`, `Expr`, `MatchClause` and `PolyExpr`
- The mutual recursion is obtained by recursive first-class module of type `TypeCheckFix.tcfix` created in `TypeInference.Main`
- Moved extraction of type hints to separate module

Module `TypeInference.Def` is left not cleaned too much, because I'll rewrite large part of it during work on #69


forell (approved):
I agree that the separation into multiple files is an improvement. The added noise involving `TCFix` seems like a small price to pay for that.

PR #88, opened 2024-05-05
Changing module names to allow use of dune utop command

Foxinio:
I think it's useful and noninvasive change, will come in hand. There may be better names to choose, but i don't think this matters that much.


ppolesiuk (commented):
Maybe `DblConfig` and `DblParser` as better names? I also suggest to change name of `Parser` directory to match the name of the module.

PR #89, opened 2024-05-06
Command line parameters for search paths

### LINKED ISSUE #85: Command line parameters for source/library search directories
List of search directories (`Config.lib_search_dirs` and `Config.local_search_dirs`) are supported by the implementation, but cannot be manually set by the user. We could add `-I` and `-L` command-line flags for setting them.

Bohun9:
This commit should resolve #85.

I am not sure if it is possible have functionality of specifying multiple arguments to a single option, like ```dbl -I path1 path2 source.dbl``` with Arg API easily.


src/dbl.ml
```diff
@@ -28,7 +28,15 @@ let cmd_args_options = Arg.align
 
     "-verbose-internal-errors",
     Arg.Set InterpLib.InternalError.verbose,
-    " Make internal errors more verbose (for debugging only)"
+    " Make internal errors more verbose (for debugging only)";
+
+    "-I",
+    Arg.String (fun p -> Config.lib_search_dirs := !Config.lib_search_dirs @ [p]),
```
forell, line 34: Expressions like `xs @ [x]` performed in a loop tend to look a bit fishy (though, to be fair, these lists are unlikely to be long). Besides, I'm not sure that we actually want to add directories from command-line options at the very end; perhaps they should take precedence over some other potential sources (_e.g.,_ environment variables), even if presently it makes no difference. I think you can solve both of these minor issues by accumulating these directories in reverse order into a temporary list, and `List.rev_append`ing them with the list in `Config` after parsing all arguments.

Also, this line is too long.

src/dbl.ml
```diff
@@ -28,7 +28,15 @@ let cmd_args_options = Arg.align
...
+    " Standard library paths";
```
forell, line 35: This option takes just one path, which I don't think is clear from the description. I'd also prefer it to be phrased in terms of an action that the option will perform, like "add a directory ...", to be consistent with the other help messages.

src/dbl.ml
```diff
@@ -28,7 +28,15 @@ let cmd_args_options = Arg.align
...
+
+    "-L",
+    Arg.String (fun p -> Config.local_search_dirs := !Config.local_search_dirs @ [p]),
+    " Local search paths";
```
forell, line 39: Same feedback as above.

forell (changes requested):
I'd like to suggest some small changes before merging this pull request. I think you should swap the `-I` and `-L` options so that `-I` configures `local_search_dirs` and `-L` configures `lib_search_dirs`. While the search paths don't accurately correspond to the traditional flags from C-related infrastructure, for now I believe this way will be more intuitive, unless you can convince me otherwise.

By the way, using `-I` or `-L` multiple times for every directory seems fine, and space-separated arguments to these options would clash with how all anonymous arguments after the filename are treated as arguments for the interpreted program anyway. So, I think this part is good as is.

Bohun9:
I agree that swapping the meaning of the flags is a good idea. I added changes you suggested.

forell (approved):
Looks good, thanks!

Issue #83, opened 2024-05-05
Type-errors in modules imported in REPL could lead to infinite loop


ppolesiuk:
The buggy behavior can be reproduced, by the two following files.
`Good.dbl`:
```
(* empty *)
```
`Bad.dbl`:
```
import Good
// some type error:
let _ = () ()
```
When in REPL we import bad, we would fall into infinite loop.
```
> import Bad ;;
./Bad.dbl:3:9-10: fatal error: This expression has type Unit. It is not a function and cannot be applied
./Bad.dbl:3:9-10: fatal error: This expression has type Unit. It is not a function and cannot be applied
./Bad.dbl:3:9-10: fatal error: This expression has type Unit. It is not a function and cannot be applied
...
```

As suggested by @forell in #80, resolving the issue #82 could help to fix this bug.

PR #90, opened 2024-05-06
REPL grouping works if there is no error

### LINKED ISSUE #82: Grouping definitions in REPL
Definitions in REPL could be grouped together. If one of them fail to type-check, the input should have no effect. The example session could look like the following.
```
> let id x = x
let foo = 42 ;;
> id foo ;;
: Int
= 42
> let bar = 13 ;;
let baz = () () (* type-error *)
<stdin>:2:11-12: fatal error: This expression has type Unit. It is not a function and cannot be applied
> bar ;;
<stdin>:1:1-3: fatal error: Unbound variable bar
>
```

Bohun9:
This commit should resolve #82.

I have made minimal changes to make it work when there are no type errors. There is a problem with infinite loop:
```
> let x = () ()
  let y = () ;;
<stdin>:1:9-10: fatal error: This expression has type Unit. It is not a function and cannot be applied
<stdin>:1:9-10: fatal error: This expression has type Unit. It is not a function and cannot be applied
<stdin>:1:9-10: fatal error: This expression has type Unit. It is not a function and cannot be applied
<stdin>:1:9-10: fatal error: This expression has type Unit. It is not a function and cannot be applied
...
```
But the other order works:
```
> let x = ()
let y = () () ;;
<stdin>:2:9-10: fatal error: This expression has type Unit. It is not a function and cannot be applied
> x ;;
<stdin>:1:1: fatal error: Unbound variable x
```
Changes in Parser/Main.ml have effect of inserting semicolons between lets and I guess that lazy handling of REPL definitions takes all those lets in one go. But I don't understand the codebase enough to tell why the loop error occurs.


src/Parser/Main.ml
```diff
@@ -34,8 +34,8 @@ let rec repl_seq imported () =
     let def = make_nowhere (Lang.Surface.DReplExpr(Desugar.tr_expr e)) in
     Seq.Cons(def, repl_seq imported)
 
-  | Raw.REPL_Def def ->
-    Seq.Cons(Desugar.tr_def def, repl_seq imported)
+  | Raw.REPL_Defs defs ->
+      (List.fold_left (fun ac def () -> Seq.Cons(Desugar.tr_def def, ac)) (repl_seq imported) defs) ()
```
ppolesiuk, line 38: Using `fold_left` for appending lists naturally gives us `rev_append` instead of `append`. The same problem occurs here: definitions are reversed during parsing. This can be exhibited by the following input to REPL.
```
> let x = ()
let y = x ;;
<stdin>:2:9: fatal error: Unbound variable x
>
```
On the other hand, the following input work, but it shouldn't.
```
> let y = x
let x = () ;;
> 
```

BTW, we try to keep source in good old 80-column style. This line is too long.

ppolesiuk (changes requested):
The order of definitions is reversed (see detailed comments).

Moreover, the problem of looping REPL (bug mentioned in the PR comment, and #83) probably comes from the fact, that single user input is not treated as a monolith *somewhere in the pipeline*. If some definition is split into two definitions `fun () -> Cons(def1, fun () -> Conf(def2, ...))` and checking of `def1` succeeds, but `def2` fails, then the evaluator (see `Eval.eval_repl`) calls the failing thunk again, which in this scenario deterministically returns `def2` instead of asking the user for another input. Therefore, thunks describing user input in REPL must be *never* split into smaller ones. In order to achieve this, the group of definitions should be propagated to type-checker and checked as a monolith. Fortunately, we already have the infrastructure for that (we can use `TypeInference.Def.check_defs` instead of `TypeInferece.Def.check_def`).

Bohun9:
I made suggested change and it seems to work.

src/Lang/Surface.ml
```diff
@@ -301,6 +301,9 @@ and def_data =
     (** Print type, evaluate, and print the expression, provided by a user in
       REPL. *)
 
+  | DReplDefs of def list
+    (** Group of REPL definitions *)
```
ppolesiuk, line 305: Why do you add a new constructor? It is easier to modify `ERepl` to accept list of definitions.

ppolesiuk (changes requested):
Adding new constructor, just to slightly modify existing functionality is a bit overkill. Please modify existing `Surface.ERepl` constructor.

Bohun9:
New changes were added.

src/Lang/Surface.ml
```diff
@@ -232,7 +232,7 @@ and expr_data =
   | EAnnot of expr * type_expr
     (** Type annotation *)
 
-  | ERepl of def Seq.t
+  | ERepl of def list Seq.t
     (** REPL. It is a lazy sequence of definitions provided by a user. *)
```
ppolesiuk, line 236: A documentation comment could be updated here.

src/Parser/Main.ml
```diff
@@ -32,14 +32,15 @@ let rec repl_seq imported () =
...
-  | Raw.REPL_Def def ->
-    Seq.Cons(Desugar.tr_def def, repl_seq imported)
+  | Raw.REPL_Defs defs ->
+      let defs = List.map Desugar.tr_def defs in
+      Seq.Cons(defs, repl_seq imported)
 
   | Raw.REPL_Import import ->
     let imported, defs = Import.import_one imported import in
-    Seq.append (List.to_seq defs) (repl_seq imported) ()
+    Seq.append (List.to_seq (List.map (fun x -> [x]) defs)) (repl_seq imported) ()
```
ppolesiuk, line 43: Getting rid of `Seq.append` and changing it to single monolithic `Seq.Cons` hopefully fix #83. If it's a change of single line (that should be modified anyway) we should try.

ppolesiuk (approved):
I approve the changes, but documentation comments need to be updated. Since there are merge conflicts with the upstream, that I'm going to resolve, I will fix these comments myself.

ppolesiuk:
Updated version fixes #83

PR #91, opened 2024-05-06
Warning about abstr modifier in pub rec block

### LINKED ISSUE #71: Warn about useless `abstr` modifier
Within a recursive block marked as `pub`, the `abstr` modifier on a data type has no effect, since the public visibility is propagated to the constructors regardless. This should probably trigger a warning, as it might not be the programmer's intention.
```
module M
  pub rec
    abstr data T = C
  end
end
```

Bohun9:
This pull request should resolve #71.


src/Parser/Desugar.ml
```diff
@@ -708,7 +708,11 @@ and tr_def ?(public=false) (def : Raw.def) =
     let (pub_type, pub_ctors) =
       match vis with
       | DV_Private  -> (public, public)
-      | DV_Abstract -> (true,   public)
+      | DV_Abstract -> if public then
```
ppolesiuk, line 711: Code formatting could be improved here. If a body of a match clause does not fit into one line, we break the line after arrow `->`. Moreover, the expression after `then` should have bigger indentation.

src/Parser/Desugar.ml
```diff
@@ -708,7 +708,11 @@ and tr_def ?(public=false) (def : Raw.def) =
...
+                       (InterpLib.Error.report ~pos:def.pos
```
ppolesiuk, line 712: `InterpLib.Error.report` is too low-level to use it here. We should use `Error.warn` and keep all message strings in one place (`Error.ml`). Of course, a new kind of warning should be added to `Error` module.

src/Parser/Desugar.ml
```diff
@@ -708,7 +708,11 @@ and tr_def ?(public=false) (def : Raw.def) =
...
+                       ~cls:InterpLib.Error.Warning
+                       "'abstr' data modifier inside public rec block");
```
ppolesiuk, line 714: The message string should be improved, because of the two following reasons.
1. It doesn't help the programmer, unless he knows the language very well.
2. It would scale poorly to nested sections (grouping several definitions that share local `implicit` declarations).

My suggestion is to say something like "This 'abstr' data modifier has no effect. It is overridden by 'public' modifier of entire group of definitions."

ppolesiuk (changes requested):
I have minor comments, that should be addressed.

Bohun9:
New changes were added.

ppolesiuk (approved):
Now, I'm happy with this changes. I'm going to merge them into master. Thanks!

PR #94, opened 2024-05-08
Implemented records and automatic accessor methods generation

### LINKED ISSUE #23: Generating methods for accessing fields of records
For record-like types, e.g.,
```
data Vec X = Vec { x : X, y : X }
```
we could automatically generate methods for accessing fields of these types. In this example it would be
```
method x {self = Vec { x }} = x
method y {self = Vec { y }} = y
```
There are some design decisions that should be made. First of all, when should we generate such methods? The possible options I see are the following (I prefer option 2 or 4).
1. Always, for ADT with single constructor (with some limitation related to existential types: see below).
2. For types explicitly marked as records (e.g., by a special keyword, or when the constructor name is omitted or the same as the name of type).
3. For each field that is shared between all constructors.
4. By explicitly marking the fields, e.g., by something like `deriving` in Haskell.

Moreover, for types that stores existential types, e.g.,
```
data T = T { X, a : X, b : Int }
```
it is not possible to generate some methods (method `a` in this example, but method `b` is still reasonable). If we follow option 2, then we should forbid existential types in records, or omit some automatically generated methods.

Foxinio:
As discussed it happens at stage of desugaring, methods are added with new variable names generated with names impossible to be entered by user.

Inside records existential types are not allowed, so if detected fatal error is thrown.

Added adequate tests

Fixes #23


src/DblParser/Main.ml
```diff
@@ -35,7 +35,8 @@ let rec repl_seq imported () =
     Seq.Cons(def, repl_seq imported)
 
   | Raw.REPL_Def def ->
-    Seq.Cons(Desugar.tr_def def, repl_seq imported)
+    let defs = Desugar.tr_def def |> List.to_seq in
+    (Seq.append defs (repl_seq imported)) ()
```
ppolesiuk, line 39: You should be careful with this line, when fixing merge conflicts. It should not provide a bug, but may brake the invariant mentioned in https://github.com/fram-lang/dbl/pull/90#pullrequestreview-2041218888, which protects us from bugs similar to #83.

src/DblParser/Desugar.ml
```diff
@@ -660,17 +666,17 @@ and tr_def ?(public=false) (def : Raw.def) =
...
-      make (DLetId(id, tr_expr e))
+        make (DLetId(id, tr_expr e))
     | LP_Fun(id, targs, iargs, args) ->
-      make (DLetFun(id, targs, iargs, tr_function args (tr_expr e)))
+        make (DLetFun(id, targs, iargs, tr_function args (tr_expr e)))
     | LP_Pat p ->
-      make (DLetPat(p, tr_expr e))
-    end
+        make (DLetPat(p, tr_expr e))
+    ]
```
ppolesiuk, line 676: I'm not sure if I like this style of code formatting. Please consider leaving `begin match .. end` and putting braces around `make (...)`. Or at least indent match clauses to align them with `match` keyword. Same comment applies to some other changes.

src/DblParser/Desugar.ml
```diff
@@ -816,4 +837,45 @@ and make_handle ~pos lbl_opt eff_opt pat body hcs =
...
+  let create_mapping i arg =
+    let old_name, new_name =
+      match (snd arg.data).data with
+      | TA_Var(name, _) -> name, name ^ "#TA_Var#" ^ string_of_int i
+      | TA_Effect -> "", "#TA_Effect#" ^ string_of_int i
+      | TA_Wildcard -> "", "#TA_Wildcard#" ^ string_of_int i
+    in
+    make (TNVar old_name, make (TA_Var(new_name, make (KWildcard)))), new_name
+  in
+  let new_named_type_args, new_names = List.split (List.mapi create_mapping named_type_args) in
```
ppolesiuk, line 851: This line doesn't fit in 80 columns. Therefore, I cannot browse this code on [VT100](https://en.wikipedia.org/wiki/VT100) terminal :smile:.

src/DblParser/Desugar.ml
```diff
@@ -816,4 +837,45 @@ and make_handle ~pos lbl_opt eff_opt pat body hcs =
...
+and generate_accessor_method_pattern named_type_args type_name =
+  let make data = { data; pos=Position.nowhere } in
+  let create_mapping i arg =
+    let old_name, new_name =
+      match (snd arg.data).data with
+      | TA_Var(name, _) -> name, name ^ "#TA_Var#" ^ string_of_int i
+      | TA_Effect -> "", "#TA_Effect#" ^ string_of_int i
+      | TA_Wildcard -> "", "#TA_Wildcard#" ^ string_of_int i
+    in
+    make (TNVar old_name, make (TA_Var(new_name, make (KWildcard)))), new_name
```
ppolesiuk, line 849: For `TA_Effect` and `TA_Wildcard` you generate a type parameter named `""` (`TN_Var ""`), which looks suspicious. I think, it should be `TNEffect` for `TA_Effect` and `TNAnon` for `TA_Wildcard`.

src/DblParser/Desugar.ml
```diff
@@ -816,4 +837,45 @@ and make_handle ~pos lbl_opt eff_opt pat body hcs =
...
+    } in
+  let pattern_gen field =
+    ArgAnnot(make (PCtor(
+      make (NPName type_name),
+      [],
+      [make (NVar field, make (PId(IdVar(false, field))))],
+      [])), type_annot) in
+  new_named_type_args, pattern_gen
+
+and create_accessor_method ~public named_type_args pattern_gen scheme =
```
ppolesiuk, line 869: This and the previous function generate some piece of code, which is hard to read directly in the abstract syntax. Maybe a short documentation comment with a simple example using concrete syntax would help.

ppolesiuk (changes requested):
Generally, these changes looks good and should be finally merged into upstream. Before that, however, some minor issues described in the detailed comments should be addressed.

src/DblParser/Main.ml
```diff
@@ -38,8 +38,11 @@ and repl_seq_main imported () =
     Seq.Cons([def], repl_seq imported)
 
   | Raw.REPL_Defs defs ->
-      let defs = List.map Desugar.tr_def defs in
+      let defs = Desugar.tr_defs defs in
       Seq.Cons(defs, repl_seq imported)
+  (* | Raw.REPL_Def def -> *)
+  (*   let defs = Desugar.tr_def def |> List.to_seq in *)
+  (*   (Seq.append defs (repl_seq imported)) () *)
```
ppolesiuk, line 45: Unnecessary commented-out code.

ppolesiuk (approved):
Now, I'm happy with these changes. I'll resolve the conflicts and merge these changes to the upstream.

PR #99, opened 2024-05-13
Fix parsing of nested arrows with effects

### LINKED ISSUE #96: Parser incorrectly recognizes arrow effect as an aplication in nested arrows
The problem can be reproduced by the following code.
```
let foo (f : Unit ->[] Unit ->[] Unit) = f () ()
```
We get the following error.
```
Err.dbl:1:21-22: fatal error: This type has kind effrow, and cannot be applied
```

Bohun9:
This pull request should resolve #96.

### DIFF
--- a/src/DblParser/Desugar.ml
+++ b/src/DblParser/Desugar.ml
@@ -164,7 +164,7 @@ and erase_eff_type (tp : Raw.type_expr) =
   | TApp({ data = TEffect _; _ }, tp) -> tp 
   | TApp(tp1, tp2) ->
       let tp1 = (erase_eff_type tp1) in
-      { data = TApp(tp1, tp2) ; pos = Position.join tp1.pos tp2.pos }
+      { pos = Position.join tp1.pos tp2.pos; data = TApp(tp1, tp2) }
   | _ -> tp
 
 and tr_scheme_expr (tp : Raw.type_expr) =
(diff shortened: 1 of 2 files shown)


ppolesiuk (changes requested):
Adding the `erase_eff_type` function seems to be an overkill: the `tr_eff_type` returns a pair, where the second component is a type without the effect. However, I see the problem: in case of arrow we want to detach effect from a scheme instead of a type. I'm afraid that your solution would not work for arrows with scheme on the RHS (I didn't check, if this claim is true), for instance `Unit ->[] ({A} -> A) -> Unit`.

A better solution, that comes to my mind is to write a version of `tr_eff_type` function (lets call it `try_detach_eff`) which works similar to `tr_eff_type`, but do not translate the type. This function should have the signature: `Raw.type_expr -> (Raw.type_expr * Raw.type_expr) option`. Using this function, we can implement `tr_eff_type` the following way.
```
let tr_eff_type tp =
  match try_detach_eff tp with
  | None -> (None, tr_type_expr tp)
  | Some(eff, tp) -> (Some (tr_type_expr eff), tr_type_expr tp)
```

Bohun9:
I made the change you suggested.

ppolesiuk (approved):
Now, the code looks reasonable. Thanks.

Issue #28, opened 2024-03-27
Optional parameters


ppolesiuk:
To the mechanism of named/implicit parameters we could add optional parameters.

PR #103, opened 2024-05-15
Add some missing functions to module List

fatmage:
Implemented most of the missing List functions (in comparison with OCaml standard library).


lib/List.dbl
```diff
@@ -1,33 +1,158 @@
...
+  | x :: xs' => if p x then drop_while p xs' else xs 
+  end 
+
 pub let rec iter f xs =
   match xs with
   | []      => ()
   | x :: xs => let () = f x in iter f xs
   end
 
+pub let rec iteri (i : Int) f xs =
```
ppolesiuk, line 150: Should `iteri` take two instead of three parameters? Of course, its implementation needs this additional accumulator, but should it be part of functions signature? Optional parameters (#28) would help here, but they are not implemented yet.

lib/List.dbl
```diff
@@ -1,33 +1,158 @@
...
+  match xs with 
+  | []      => ()
+  | x :: xs => let _ = f i x in iteri (i + 1) f xs
+  end 
+
 pub let iter2 {`re : {type X} -> Unit ->[|_] X} =
```
ppolesiuk, line 156: Maybe `` `re`` could be monomorphic? It would allow to not raise an error in case of lists of different lengths.

lib/List.dbl
```diff
@@ -36,3 +161,95 @@ pub let iter2 {`re : {type X} -> Unit ->[|_] X} =
     | _                => `re ()
   end
   in iter
+
+pub let init (len : Int) f = 
+  let rec init_aux (i : Int) l f = if i > l then [] else (f i) :: init_aux (i+1) l f
```
ppolesiuk, line 166: Firstly, this line is too long. It should fit on 80 column [VT100](https://en.wikipedia.org/wiki/VT100) terminal.

Secondly, The `l` parameter of `init_aux` is not necessary. We can use `len` instead.

lib/List.dbl
```diff
@@ -36,3 +161,95 @@ pub let iter2 {`re : {type X} -> Unit ->[|_] X} =
...
+  match xs with
+  | []      => acc 
+  | x :: xs => fold_left f (f acc x) xs 
+  end 
+
+pub let rec fold_right f xs acc =
+  match xs with 
+  | []      => acc 
+  | x :: xs => f x (fold_right f xs acc)
+  end
```
ppolesiuk, line 179: Should we keep the order of parameters in `fold_left`/`fold_right` the same as in OCaml, or just use convention from Haskell? I prefer OCaml style, but we can discuss it.

lib/List.dbl
```diff
@@ -36,3 +161,95 @@ pub let iter2 {`re : {type X} -> Unit ->[|_] X} =
...
+  | x :: xs => p x && for_all p xs
+  end 
+
+pub let rec exists p xs =
+  match xs with 
+  | []      => False 
+  | x :: xs => p x || exists p xs 
+  end
+
+pub let find {`re : {type X} -> Unit ->[|_] X} p xs =
```
ppolesiuk, line 193: Again, should `` `re`` be monomorphic?

lib/List.dbl
```diff
@@ -36,3 +161,95 @@ pub let iter2 {`re : {type X} -> Unit ->[|_] X} =
...
+
+pub let rec find_index_opt p xs = 
+  let rec find_index_opt_aux p (i : Int) xs =
+    match xs with 
+    | []      => None 
+    | x :: xs => if p x then Some i else find_index_opt_aux p (i + 1) xs
+    end
+  in find_index_opt_aux p 0 xs
+
+pub let find_map {`re : {type X} -> Unit ->[|_] X} f xs =
```
ppolesiuk, line 215: Same question as above.

lib/List.dbl
```diff
@@ -36,3 +161,95 @@ pub let iter2 {`re : {type X} -> Unit ->[|_] X} =
...
+    | []      => None 
+    | x :: xs => 
+      match f i x with 
+      | Some x' => Some x'
+      | None    => find_mapi_opt_aux f (i + 1) xs 
+      end 
+    end 
+  in find_mapi_opt_aux f 0 xs
+
+pub let find_all = filter
```
ppolesiuk, line 249: Do we need both names for the same function (`find_all` and `filter`)?

lib/List.dbl
```diff
@@ -36,3 +161,95 @@ pub let iter2 {`re : {type X} -> Unit ->[|_] X} =
...
+    end 
+  in find_map_aux f xs
+
+pub let rec find_map_opt f xs =
+  match xs with 
+  | []      => None 
+  | x :: xs => 
+    match f x with 
+    | Some x => Some x
+    | None    => find_map_opt f xs
```
ppolesiuk, line 233: Code formatting could be improved in this line

lib/List.dbl
```diff
@@ -36,3 +161,95 @@ pub let iter2 {`re : {type X} -> Unit ->[|_] X} =
...
+  match xs with 
+  | []      => None 
+  | x :: xs => 
+    match f x with 
+    | Some x => Some x
+    | None    => find_map_opt f xs 
+    end 
+  end
+
+pub let rec find_mapi_opt f xs =
```
ppolesiuk, line 237: There is an asymmetry between `find_map` and `find_mapi` functions (the second have only `*_opt` version). Maybe, if they take a parameter that returns `Option`, we could have only the two functions: `find_map` and `find_mapi`, both returning `Option.

examples/Prolog.dbl
ppolesiuk, line 1: I don't see a useful family of function here: `fold_{left/right}_map` and `fold_{left/right}_{filter/concat}_map` (not all of them are available in OCaml, but I found them very useful). Maybe we can consider get rid of `fold_right_*` version, as I've never use it.

Moreover, we can consider adding method versions of these functions, e.g.,

```
pub method map f = map f self
pub method add ys = append self ys
pub method filter f = filter f self
...
```

ppolesiuk (commented):
Thanks for great work. I'm not going to merge these changes yet, since there are some design decisions that should be discussed first (and there are some minor code formatting issues). Moreover, should we start writing documentation comments in stdlib? I know, we don't have `framdoc` tool yet, but this should not prevent us from starting writing documentation.

examples/Prolog.dbl
fatmage: What should be the behaviour of `fold_{left/right}_{filter/concat}_map`?

Should we add method versions for all of these functions or only for some of them?
I think we could add them for all functions, but if not then which ones should be included?

ppolesiuk: > What should be the behaviour of `fold_{left/right}_{filter/concat}_map`?

There are `fold_left_map` and `filter_map` in OCaml. Just extrapolate.

> Should we add method versions for all of these functions or only for some of them?

I think, we can add all of them.

lib/List.dbl
```diff
@@ -1,38 +1,428 @@
...
+pub method find_mapi_err { `onError } f = find_mapi_err f self
+pub method fold_left_map f acc = fold_left_map f acc self
+pub method fold_left_filter_map f acc = fold_left_filter_map f acc self
+pub method fold_left_concat_map f acc = fold_left_concat_map f acc self
+pub method mem { E, method equal : E -> E ->[] Bool } v =
+  mem {E = E} v self
+pub method equal { E, method equal : E -> E ->[] Bool } xs =
+  equal {E = E} self xs
+pub method neq { E, method equal : E -> E ->[] Bool } xs =
+  not (equal {E = E} self xs)
```
ppolesiuk, line 428: No new line at end of file.

ppolesiuk (changes requested):
The implementation looks reasonable, but we agreed to use `camelCase` naming convention. After this change, I would merge this PR into upstream.

PR #104, opened 2024-05-17
Builtin type Char

wojpok:
Commit contains fully working implementation of Chars and some useful builtin functions (compares, toString and conversion between char list and string type).


lib/Prelude.dbl
```diff
@@ -87,6 +87,18 @@ pub method makeString {`re : {type X} -> Unit ->[|_] X, self : String}
     (extern dbl_strMake : Int -> String) n
   else `re ()
 
+pub method toList     = (extern dbl_strToListChr : String -> List Char) self
```
ppolesiuk, line 90: Should this function be implemented as extern? I think it's expressible using other functions.

My concerns come from the principle to keep extern functions as simple as possible. With simpler set of extern functions it would be easier to add new backend for the interpreter/compiler.

lib/Prelude.dbl
```diff
@@ -87,6 +87,18 @@ pub method makeString {`re : {type X} -> Unit ->[|_] X, self : String}
...
+pub let charListToStr = (extern dbl_chrListToStr : List Char -> String)
```
ppolesiuk, line 91: As above, this function is quite complex. Can we implement it using simpler means?

src/DblParser/Lexer.mll
```diff
@@ -142,6 +142,8 @@ rule token = parse
   | lid_start var_char* as x { tokenize_ident x }
   | uid_start var_char* as x { YaccParser.UID x }
   | '`' lid_start var_char* as x { YaccParser.TLID x }
+  | '\'' (_ as ch) '\''   { YaccParser.CHR ch }
+  | "\'\\" (_ as ch) '\'' { YaccParser.CHR (match ch with 'n' -> '\n' | 't' -> '\t' | '\\' -> '\\' | _ -> failwith "Invalid char escape")  }
```
ppolesiuk, line 146: I have three comments regarding this single line:

1. In lexing of string literals we do something for escaping sequences. We should not duplicate the code and share this "something" between characters and strings.
2. Raising errors by `failwith "some string without pointing the position in the source"` is very bad. We should use `Error` module. See the reporting invalid escape codes in string literals.
3. This line is too long. We try to keep the source code in 80 columns.

src/Eval.ml
```diff
@@ -88,8 +110,17 @@ let extern_map =
...
+    "dbl_strGet",  str_fun (fun s -> int_fun (fun n -> VChr s.[n]));
     "dbl_strMake", int_fun (fun n -> VStr (String.make 1 (Char.chr n)));
+    "dbl_eqChr",   chr_cmpop ( = );
+    "dbl_neqChr",  chr_cmpop ( <> );
+    "dbl_gtChr",   chr_cmpop ( > );
+    "dbl_ltChr",   chr_cmpop ( < );
+    "dbl_geChr",   chr_cmpop ( >= );
+    "dbl_leChr",   chr_cmpop ( <= );
+    "dbl_chrToString",  chr_fun (fun c -> VStr (Char.escaped c));
+    "dbl_strToListChr", str_fun (fun s -> String.to_seq s |> List.of_seq |> List.map (fun c -> VChr c) |> of_list);
```
ppolesiuk, line 122: This line is too long.

ppolesiuk (changes requested):
Thanks for your work. I'm mostly very happy with these changes, but I have some concerns regarding the implementation of a lexer. Moreover, I think some design decisions should be discussed before merging these PR. Namely, the following.

1. The design of char-related extern functions (see comments below)
2. Should `Char` be a separate builtin type for the whole stack of intermediate languages? While this is a reasonable decision for Surface, but we can consider merging Char and Int in Core. The main possible benefit is a simplification these Core and Untyped and reducing the set of necessary extern functions.

The other issue is that I don't see the two very useful functions/method in the standard library: `code` and `chr` for converting between Char and Int.

lib/Prelude.dbl
wojpok: ```toList``` could be made using strings' ```get``` method and ```charListToStr``` using string concatenation,  but I am really worried about performance of said approach.

ppolesiuk: In `toList` using `get` will be similar to the current implementation, but in different language. There should be no performance issues here.

As for `charListToStr`, using concatenation would have logarithmic performance overhead if implemented properly. Maybe it is not a big overhead, but for me it may justify to implement it as extern. However, if we add to the library mutable and freezable arrays (I talked about them with @Foxinio, and I'm going to add an issue that explain our idea) then we would be able to provide efficient and simple implementation of `charListToString`. As for now, I'm happy with current implementation.

wojpok:
Overall I fixed lexer and improved prelude. Chars are partially erased from Core: now builtin Int comparators can be used for Chars as well.

test/ok/ok0101_chars.dbl
```diff
@@ -0,0 +1,9 @@
+let _ = 'a'
+let _ = '''
+let _ = '\n'
+let _ = '\t'
+let _ = '\\'
+
+let foo (x : Char) = x
+
+let _ = foo '1'
```
ppolesiuk, line 9: No new line at end of file

ppolesiuk (commented):
As far as I understand, characters become integers in the evaluator. What is the reason for not merging these two types earlier, I mean in `ToCore`?

PR #105, opened 2024-05-17
Recursive methods and better polymorphic recursion

### LINKED ISSUE #70: Recursive methods
As it stands recursive methods are not usable. The type guessed for the recursive definition before it's type-checked does not contain enough information to add the method to the environment.

ppolesiuk:
Now, guessing the type of a recursive value looks into lambda abstractions and type annotations inside a value in order to gather more information about the type.

Fixes #70


forell (approved):
Looks like a solid improvement with no downsides as far as I can see.

PR #106, opened 2024-05-18
Types and values attached to delimiter in Core

### LINKED ISSUE #101: Attaching types and values to delimiters in Core
Attaching types to delimiters (like in [this paper](https://doi.org/10.4230/LIPIcs.FSCD.2019.30)) can be generalized to attaching types and values. This in turn leads to simple encoding of dynamic effect handlers (similar to the one presented in the mentioned paper, but simpler: with attached values we can get rid of encoding of (non-mutable) state that remembers the handler). This feature should be added to Core, because it allows us to add dynamic handlers at low cost.

ppolesiuk:
Added types and values attached to delimiter in Core and Untyped intermediate languages. These features are not used by `ToCore` translation, yet.

I've implemented them slightly differently than the proposed approach during the discussion with @forell. When we separate label signatures (with irrelevant proofs similar to the ones in ADT) from labels, as we discussed, we land in inconvenient situation in Core when labels are mutually recursive with signatures: label definition (which lives on the level of types) depends on value component of signautre (the proof). This in turn makes type-checking more complicated.

Therefore, I decided to simply enrich label type. I hope, the proposed solution for Surface/Unif can still be implemented, by translating effect signatures to ADTs with single label fields in Core, and translating signature constraints to data constraints.

Fixes #101

### DIFF
--- a/src/Eval.ml
+++ b/src/Eval.ml
@@ -27,6 +27,7 @@ and 'v comp = ('v -> ans) -> ans
 (** Stack frame (related to control operators) *)
 and frame =
   { f_label : UID.t
+  ; f_vals  : value list
   ; f_ret   : (value -> value comp)
   ; f_cont  : (value -> ans)
   }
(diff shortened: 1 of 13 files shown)


forell (approved):
The implementation looks reasonable and I agree with the argument for enriching the built-in label type in Core.

PR #112, opened 2024-05-31
Rename DBL language to Fram

### LINKED ISSUE #93: Rename DBL language to Fram
Since we are going to bootstrap Fram compiler, the language should be named "Fram", and should use `.fram` file extension. We will still use DBL as the name of this interpreter. The following things should be done.

- [ ] Change the default file extension from `.dbl` to `.fram`.
- [ ] Rename all library, test, and example files.
- [ ] Update comments in examples.
- [ ] Update `README.md`.
- [ ] Update repository description.
- [ ] Make sure that all uses of `DBL` refers to the interpreter, but not to the language.

ppolesiuk:
DBL name still refers to the implementation of interpreter.

Fixes #93


forell (approved):
Looks good, thanks!

PR #113, opened 2024-06-03
Better position printing in error messages

Foxinio:
Added better position printing

Makes it so that instead of printing line and character number
we print piece of code specified by position of an error

Before:
```
$ dbl test/err/tc_0005_specialBinops.dbl
test/err/tc_0005_specialBinops.dbl:1:9-10: error: This expression has
type String, but an expression was expected of type Bool
test/err/tc_0005_specialBinops.dbl:2:9-10: error: This expression has
type String, but an expression was expected of type Bool
test/err/tc_0005_specialBinops.dbl:2:15-16: error: This expression has
type String, but an expression was expected of type Bool
test/err/tc_0005_specialBinops.dbl:2:21-22: error: This expression has
type String, but an expression was expected of type Bool
test/err/tc_0005_specialBinops.dbl:3:17: error: This expression has type
Int, but an expression was expected of type Bool
test/err/tc_0005_specialBinops.dbl:4:9-10: error: This expression has
type String, but an expression was expected of type Unit
```

Now:
```
$ dbl test/err/tc_0005_specialBinops.dbl
error: This expression has type String, but an expression was expected of type Bool
-> test/err/tc_0005_specialBinops.dbl
 1 | let _ = "" && True
   |         ^^
 2 | let _ = "" && "" || ""
 3 | let _ = True || 5
error: This expression has type String, but an expression was expected of type Bool
-> test/err/tc_0005_specialBinops.dbl
 1 | let _ = "" && True
 2 | let _ = "" && "" || ""
   |         ^^
 3 | let _ = True || 5
error: This expression has type String, but an expression was expected of type Bool
-> test/err/tc_0005_specialBinops.dbl
 1 | let _ = "" && True
 2 | let _ = "" && "" || ""
   |               ^^
 3 | let _ = True || 5
error: This expression has type String, but an expression was expected of type Bool
-> test/err/tc_0005_specialBinops.dbl
 1 | let _ = "" && True
 2 | let _ = "" && "" || ""
   |                     ^^
 3 | let _ = True || 5
error: This expression has type Int, but an expression was expected of type Bool
-> test/err/tc_0005_specialBinops.dbl
 2 | let _ = "" && "" || ""
 3 | let _ = True || 5
   |                 ^
 4 | let _ = ""; 5
error: This expression has type String, but an expression was expected of type Unit
-> test/err/tc_0005_specialBinops.dbl
 3 | let _ = True || 5
 4 | let _ = ""; 5
   |         ^^

```

I think this is a good quality of life improvement.

### DIFF
--- a/src/InterpLib/Error.ml
+++ b/src/InterpLib/Error.ml
@@ -29,12 +29,12 @@ let report ?pos ~cls msg =
     | Warning -> "warning"
     | Note    -> "note"
   in
-  match pos with
+  match Option.bind pos Position.get_text_range with
   | None ->
     Printf.eprintf "%s: %s\n" name msg
   | Some pos ->
     Printf.eprintf "%s: %s\n%s\n"
-      name msg (Position.get_text_range pos)
+      name msg pos
 
 let assert_no_error () =
   if !err_counter <> 0 then
--- a/src/InterpLib/Error.ml
+++ b/src/InterpLib/Error.ml
@@ -33,8 +33,8 @@ let report ?pos ~cls msg =
   | None ->
     Printf.eprintf "%s: %s\n" name msg
   | Some pos ->
-    Printf.eprintf "%s: %s: %s\n"
-      (Position.to_string pos) name msg
+    Printf.eprintf "%s: %s\n%s\n"
+      name msg (Position.get_text_range pos)
 
 let assert_no_error () =
   if !err_counter <> 0 then
--- a/src/InterpLib/Error.ml
+++ b/src/InterpLib/Error.ml
@@ -29,12 +29,15 @@ let report ?pos ~cls msg =
     | Warning -> "warning"
     | Note    -> "note"
   in
-  match Option.bind pos Position.get_text_range with
-  | None ->
-    Printf.eprintf "%s: %s\n" name msg
-  | Some pos ->
+  match pos, Option.bind pos Position.get_text_range with
+  | Some pos, None ->
+    Printf.eprintf "%s: %s: %s\n"
+      (Position.to_string pos) name msg
+  | _, Some pos ->
     Printf.eprintf "%s: %s\n%s\n"
       name msg pos
+  | None, _ ->
+    Printf.eprintf "%s: %s\n" name msg
 
 let assert_no_error () =
   if !err_counter <> 0 then
--- a/src/Utils/Position.ml
+++ b/src/Utils/Position.ml
@@ -99,91 +99,92 @@ let to_string pos =
 
 module PrettyPrinting =
 struct
+  type underline_options =
+    | NoUnderline
+    | UnderlineBegining
+    | UnderlineIfOneLine
+    | UnderlineAlways
 
-type underline_option =
-  | NoUnderline
-  | UnderlineBegining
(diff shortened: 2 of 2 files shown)


src/InterpLib/Error.ml
```diff
@@ -29,12 +29,12 @@ let report ?pos ~cls msg =
     | Warning -> "warning"
     | Note    -> "note"
   in
-  match pos with
+  match Option.bind pos Position.get_text_range with
```
ppolesiuk, line 32: I think we should not use `Option.bind` here, because we have three cases instead of two:

1. we do not want to print the position (`pos = None`)
2. we want to print the position, but we are unable to extract the text range, e.g. in REPL, but I thing it would change in the future; other example is when line directives (issue #128) are broken
3. we get the text range that we can print.

Current implementation merges the first two cases.

src/Utils/Position.ml
```diff
@@ -96,3 +96,106 @@ let to_string pos =
       pos.pos_fname
       pos.pos_start_line (start_column pos)
       pos.pos_end_line   (end_column pos)
+
+module PrettyPrinting =
+struct
+  type underline_options =
```
ppolesiuk, line 102: The meaning of types in this module should be documented.

src/Utils/Position.ml
```diff
@@ -96,3 +96,106 @@ let to_string pos =
...
+  let default_options = {
+    context = 1;
+    underline = (fun _ -> UnderlineIfOneLine);
+    add_line_numbers = true;
+  }
+
+end
+
+(** Get text from file from given position *)
+let get_text_range ?(options = PrettyPrinting.default_options) ?channel (pos : t) =
```
ppolesiuk, line 123: Currently, the pretty printer uses only default options, right? If so, what are the future plans for other possible options?

src/Utils/Position.ml
```diff
@@ -96,3 +96,106 @@ let to_string pos =
...
+  if Fun.negate Sys.file_exists pos.pos_fname then None else
+  let file_chunk = match channel with
+    | None -> In_channel.with_open_text pos.pos_fname process_file
+    | Some ch -> process_file ch
+  in
+  let pp_file_name = " -> " ^ pos.pos_fname ^ "\n" in
+  Some (pp_file_name ^ file_chunk)
+
+
+
```
ppolesiuk, line 201: Extra new lines at end of file should be removed.

ppolesiuk (changes requested):
It looks like a solid piece of work. But I would like to not give up position printing in REPL, so the first of the comments below should be addressed.

src/InterpLib/Error.ml
```diff
@@ -29,12 +29,15 @@ let report ?pos ~cls msg =
     | Warning -> "warning"
     | Note    -> "note"
   in
-  match pos with
-  | None ->
-    Printf.eprintf "%s: %s\n" name msg
-  | Some pos ->
+  match pos, Option.bind pos Position.get_text_range with
```
ppolesiuk, line 32: Maybe we can match on `Option.map Position.get_text_range pos`? Its type is more precise.

PR #114, opened 2024-06-03
Rudimentary module patterns

forell:
Adds support for pattern matching such as `let C { module M } = ...`. This hopefully brings us closer to making functors expressed as regular functions convenient to use. One limitation is that other fields can't be mixed with a module field, that is, the pattern `C { module M, x }` is currently invalid. Additionally, the `module M` syntax cannot be used to instantiate named parameters yet.

Removing the limitations would probably be best done alongside reworking how named parameter lists are treated.

### DIFF
--- a/src/TypeInference/Pattern.ml
+++ b/src/TypeInference/Pattern.ml
@@ -140,41 +140,50 @@ let get_ctor_info ~pos ~env (cpath : S.ctor_name S.path S.node) tp =
   | Whnf_Effect _ | Whnf_Effrow _ ->
     failwith "Internal kind error"
 
+(** Introduce all named types (from the constructor scheme) into the
+  environment. Returns extended environment, refreshed type parameters, and a
+  renaming used for refreshing. *)
+let open_named_types ~public env targs =
+  let open_named_type (env, sub) (name, x) =
+    let (env, y) =
+      let kind = T.TVar.kind x in
+      match name with
+      | T.TNAnon   -> Env.add_anon_tvar env kind
+      | T.TNEffect ->
+        assert (T.Kind.is_effect kind);
+        Env.add_the_effect env
+      | T.TNVar x  -> Env.add_tvar ~public env x kind
+    in
+    ((env, T.Subst.rename_to_fresh sub x y), y)
+  in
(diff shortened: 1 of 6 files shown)


src/TypeInference/Pattern.ml
```diff
@@ -140,7 +140,68 @@ let get_ctor_info ~pos ~env (cpath : S.ctor_name S.path S.node) tp =
   | Whnf_Effect _ | Whnf_Effrow _ ->
     failwith "Internal kind error"
 
-let rec check_ctor_named_args ~pos ~env ~scope nps named =
+(** Introduce all named types and named arguments into the environment.
+  The type variables are refreshed, so the function returns a renaming. *)
+let open_named ~pos ~public env targs named =
```
ppolesiuk, line 145: This function is long, and could be improved by defining two helper functions: one for types, and one for values.

src/TypeInference/Pattern.ml
```diff
@@ -140,7 +140,68 @@ let get_ctor_info ~pos ~env (cpath : S.ctor_name S.path S.node) tp =
...
+        targs ctor.ctor_targs in
+    let ctor_named = List.map (T.NamedScheme.subst sub2) ctor.ctor_named in
+    Uniqueness.check_named_pattern_uniqueness nps;
+    let nps = List.map (tr_named_pattern env) nps in
+    let (env, bn1, ps1) =
+      check_ctor_named_args ~pos ~env ~scope nps ctor_named in
+    (env, scope, sub2, tvars, ps1, bn1)
+  | S.CNModule modname ->
+    (* TODO: This may seem inconsistent with the other case as implicits aren't
+       introduced to the current namespace, just the provided module name. *)
```
ppolesiuk, line 197: Indeed. This should be discussed and we should find a general policy of implicit introduction of variables. Before that, we can live with what is done.

ppolesiuk (approved):
Since this code looks reasonable and implements functionality needed to develop the standard library I accept this changes. I have some minor comments, and I would apply necessary improvements myself. However, further development of this feature requires some major refactoring (especially on how named parameters are represented in Surface). We can do this later.

PR #116, opened 2024-06-03
Added Int64 builtin type with dedicated module in stdlib

Foxinio:
Added int64 on all stages of pipeline, used normal int as template Added standard library functionality (most of that as builtin functions)
Also added neg builtin for int as we missed it for some reason


lib/Int64.fram
ppolesiuk, line 1: I think, the methods defined in this module should be moved to Prelude or at least some module imported from Prelude. The user should not import Int64, just to write `2L + 2L`.

lib/Int64.fram
```diff
@@ -0,0 +1,71 @@
...
+pub method lt    {self : Int64} = (extern dbl_ltInt64  : Int64 -> Int64 -> Bool) self
+pub method ge    {self : Int64} = (extern dbl_geInt64  : Int64 -> Int64 -> Bool) self
+pub method le    {self : Int64} = (extern dbl_leInt64  : Int64 -> Int64 -> Bool) self
+
+pub method toString = (extern dbl_int64ToString : Int64 -> String) self
+
+pub method neg {self : Int64} = (extern dbl_negInt64 : Int64 -> Int64) self
+pub method add {self : Int64} = (extern dbl_addInt64 : Int64 -> Int64 -> Int64) self
+pub method sub {self : Int64} = (extern dbl_subInt64 : Int64 -> Int64 -> Int64) self
+pub method mul {self : Int64} = (extern dbl_mulInt64 : Int64 -> Int64 -> Int64) self
```
ppolesiuk, line 13: Some lines in this module exceeds 80 columns.

lib/Int64.fram
```diff
@@ -0,0 +1,71 @@
...
+
+pub let zero = 0L
+pub let one  = 1L
+pub let minus_one = (1L).neg
```
ppolesiuk, line 17: What is the motivation behind introducing these three definitions?

lib/Int64.fram
```diff
@@ -0,0 +1,71 @@
...
+pub method succ {self : Int64} = self + one
+pub method pred {self : Int64} = self - one
```
ppolesiuk, line 19: What is the motivation behind these two methods? Why the programmer cannot just write `x + 1L`?

lib/Int64.fram
```diff
@@ -0,0 +1,71 @@
...
+
+pub method shiftl  = (extern dbl_lslInt64 : Int64 -> Int64 -> Int64) self
+pub method shiftr  = (extern dbl_lsrInt64 : Int64 -> Int64 -> Int64) self
+pub method ashiftr = (extern dbl_asrInt64 : Int64 -> Int64 -> Int64) self
+
+pub let int64_of_int = (extern dbl_int64ToInt : Int -> Int64)
+
+
+
+
```
ppolesiuk, line 41: There are many empty lines here. Why?

lib/Int64.fram
```diff
@@ -0,0 +1,71 @@
...
+pub method land = (extern dbl_andInt64 : Int64 -> Int64 -> Int64) self
+pub method lor  = (extern dbl_orInt64  : Int64 -> Int64 -> Int64) self
+pub method lxor = (extern dbl_xorInt64 : Int64 -> Int64 -> Int64) self
+
+pub method shiftl  = (extern dbl_lslInt64 : Int64 -> Int64 -> Int64) self
+pub method shiftr  = (extern dbl_lsrInt64 : Int64 -> Int64 -> Int64) self
+pub method ashiftr = (extern dbl_asrInt64 : Int64 -> Int64 -> Int64) self
+
+pub let int64_of_int = (extern dbl_int64ToInt : Int -> Int64)
```
ppolesiuk, line 37: 1. We agreed to use camelCase naming convention.
2. Maybe we could define a method toInt64 for type Int instead of this function.

lib/Int64.fram
```diff
@@ -0,0 +1,71 @@
...
+
+
+
+
+pub method to_string {`re : {type X} -> Unit ->[|_] X, self : Int64} =
```
ppolesiuk, line 42: Why does the `to_string` (which should be named `toString`) method can raise an error?

lib/Int64.fram
```diff
@@ -0,0 +1,71 @@
...
+	if self < 0x7fffffffffffffffL then
+		(extern dbl_int64ToString : Int64 -> String) self
+	else `re ()
+
+pub method to_string_opt {self : Int64} =
```
ppolesiuk, line 47: I think this method can be removed.

lib/Int64.fram
```diff
@@ -0,0 +1,71 @@
...
+	if self < 0x7fffffffffffffffL then
+		Some ((extern dbl_int64ToString : Int64 -> String) self)
+	else None
+
+pub method to_int {`re : {type X} -> Unit ->[|_] X, self : Int64} =
```
ppolesiuk, line 52: 1. The name should be camelCase
2. The numbers are signed, not unsigned, so bound checks should look differently.

lib/Int64.fram
```diff
@@ -0,0 +1,71 @@
...
+		(extern dbl_intToInt64 : Int64 -> Int) self
+	else `re ()
+
+pub method to_int_opt {self : Int64} =
+	if self < 0x7fffffffffffffffL then
+		Some ((extern dbl_intToInt64 : Int64 -> Int) self)
+	else None
+
+pub method abs {self : Int64} =
+	if self < 0L then self.neg else self
```
ppolesiuk, line 63: Why `self.neg` instead of `- self`?

src/DblParser/Desugar.ml
ppolesiuk, line 1: This file looks OK, but some lines are longer than 80.

src/DblParser/Lexer.mll
```diff
@@ -80,13 +80,23 @@ let tokenize_ident str =
...
-  {|^\(0[bB][01]*\|0[oO][0-7]*\|[0-9]*\|0[xX][0-9a-fA-F]*\)$|}
+  {|^\(0[bB][01]*\|0[oO][0-7]*\|[0-9]*\|0[xX][0-9a-fA-F]*\)[lL]?$|}
 
 let tokenize_number pos str =
   if not (Str.string_match num_regex str 0) then
     Error.fatal (Error.invalid_number
       (Position.of_lexing (String.length str) pos)
       str)
+  else if String.length str > 0 &&
+    (str.[String.length str - 1] = 'l' ||
```
ppolesiuk, line 91: In OCaml, the `l` suffix stands for `int32` literals. Maybe we should not use `l` here for similar reason?

src/Eval.ml
```diff
@@ -68,13 +71,27 @@ let rec of_list = function
...
+let int_unop  op = int_fun  (fun x -> VNum (op x))
 let int_binop op = int2_fun (fun x y -> VNum (op x y))
 let int_cmpop op = int2_fun (fun x y -> of_bool (op x y))
 
+let int64_fun f = VFn (fun v cont ->
+  match v with
+  | VNum64 n -> cont (f n)
+  | _ -> failwith "Runtime error!")
+
+let int64'2_fun f = int64_fun (fun x -> int64_fun (f x))
```
ppolesiuk, line 83: I don't like the name of this function. Maybe `int64_fun2` would be better? Then of course, we should rename `int2_fun` to `int_fun2`.

src/Eval.ml
```diff
@@ -111,13 +128,35 @@ let extern_map =
...
+    "dbl_asrInt64",      int64_int_op Int64.shift_right;
+    "dbl_eqInt64",       int64_cmpop ( = );
+    "dbl_neqInt64",      int64_cmpop ( <> );
+    "dbl_gtInt64",       int64_cmpop ( > );
+    "dbl_ltInt64",       int64_cmpop ( < );
+    "dbl_geInt64",       int64_cmpop ( >= );
+    "dbl_leInt64",       int64_cmpop ( <= );
+    "dbl_intToInt64",    int_fun (fun n -> VNum64 (Int64.of_int n));
+    "dbl_int64ToInt",    int64_fun (fun n -> VNum (Int64.to_int n));
+    "dbl_int64ToString", int64_fun (fun n -> VStr (Int64.to_string n));
```
ppolesiuk, line 151: This list is getting longer and longer. Maybe we should consider moving it to separate file.

src/Eval.ml
```diff
@@ -111,13 +128,35 @@ let extern_map =
...
   ] |> List.to_seq |> Hashtbl.of_seq
 
 (* ========================================================================= *)
 
 let to_string (v : value) =
   match v with
   | VNum n   -> string_of_int n
+  | VNum64 n   -> Int64.to_string n
```
ppolesiuk, line 159: Code formatting could be improved here in order to keep all arrows aligned.

src/Lang/CorePriv/SExprPrinter.ml
```diff
@@ -153,6 +153,7 @@ let rec tr_expr (e : Syntax.expr) =
 and tr_value (v : Syntax.value) =
   match v with
   | VNum n -> List [ Sym (string_of_int n) ]
+  | VNum64 n -> List [ Sym (Int64.to_string n) ]
```
ppolesiuk, line 156: I think we should add `L` suffix here, to distinguish Int and Int64 literals.

src/Lang/UnifPriv/BuiltinType.ml
```diff
@@ -7,6 +7,8 @@
 (** Int type *)
 let tv_int = TVar.fresh KindBase.k_type
 
+(** Int64 type *)
+let tv_int64 = TVar.fresh KindBase.k_type
 (** String type *)
```
ppolesiuk, line 12: Empty line is missing here.

src/ToCore/Main.ml
ppolesiuk, line 1: Again, lines are too long.

src/TypeInference/RecDefs.ml
```diff
@@ -182,7 +182,7 @@ let rec guess_rec_value_type env (e : S.expr) =
   | EAnnot(_, tp) ->
     (Type.tr_ttype env tp, Impure)
 
-  | EUnit | ENum _ | EStr _ | EChr _ | EPoly _ | EApp _ | EDefs _ | EMatch _
+  | EUnit | ENum _ | ENum64 _ | EStr _ | EChr _ | EPoly _ | EApp _ | EDefs _ | EMatch _
```
ppolesiuk, line 185: This line is too long

test/test_suite
```diff
@@ -18,6 +18,15 @@ function simple_examples {
...
 run_with_flags simple_examples ""
 
+function simple_stdlib_tests {
+	for file in test/stdlib/*.dbl
+	do
+		simple_test $file
+	done
+}
+
+run_with_flags simple_stdlib_tests ""
```
ppolesiuk, line 28: I agree, we should have some tests for stdlib. Maybe more conventional unit testing would be nice in the future. But now, this should be enough.

test/stdlib/stdlib0001_Int64.dbl
ppolesiuk, line 1: We start test numbering from 0.

test/test_suite
```diff
@@ -18,6 +18,15 @@ function simple_examples {
 
 run_with_flags simple_examples ""
 
+function simple_stdlib_tests {
+	for file in test/stdlib/*.dbl
```
ppolesiuk, line 22: We have changed the default file extension to `.fram`.

ppolesiuk (changes requested):
I like the direction of this work. However, some things should be improved or at least discussed. See comments below.

lib/Int64.fram
Foxinio: because it returns error:
```
> method neg {self : Int} = - self ;;
<stdin>:1:27: fatal error: Unbound variable (- .)
```
i dont know if we have unary minus, and with how operators are implemented if we can have it along side binary one

lib/Int64.fram
Foxinio: That's true I'll add module Int64 to Prelude, and in the future when Prelude grows too big we will separate it into multiple files

lib/Int64.fram
Foxinio: as above

lib/Int64.fram
ppolesiuk: Unary minus is supported, but I don't see it in the Prelude. It can be implemented in the following way:
```
method neg {self : Int} = 0 - self
method fn (- .) = neg
```

lib/Prelude.fram
```diff
@@ -123,3 +126,66 @@ pub let printInt   = extern dbl_printInt   : Int ->[IO] Unit
...
+	pub let zero = 0L
+	pub let one  = 1L
+	pub method succ {self : Int64} = self + one
+	pub method pred {self : Int64} = self - one
+
+	pub let ofInt (n : Int) = n.toInt64
+	pub method toString {self : Int64} =
+		(extern dbl_int64ToString : Int64 -> String) self
+	pub method toInt {`re : {type X} -> Unit ->[|_] X, self : Int64} =
+		if self <= 0x3fffffffffffffffL && self >= 0x7fffffffffffffffL then
```
ppolesiuk, line 176: These bounds looks suspicious. There are no numbers that satisfy both conditions.

lib/Prelude.fram
```diff
@@ -123,3 +126,66 @@ pub let printInt   = extern dbl_printInt   : Int ->[IO] Unit
...
+		else None
+
+	pub method abs {self : Int64} =
+		if self < 0L then self.neg else self
+	pub method min {self : Int64} oth =
+		if self <= oth then self else oth
+	pub method max {self : Int64} oth =
+		if self >= oth then self else oth
+
+end
```
ppolesiuk, line 191: I would change the indentation from tabulations to double spaces in this module.

ppolesiuk (approved):
I generally approve the changes. There are two minor issues that should be fixed. You can fix them, or if not, I'll do it tomorrow during the merge of these changes. BTW, this PR does two things: it adds Int64 type, and refactors the evaluator. Generally, I would prefer to have two separate PRs in such a case, but since the work is already done, you can leave it as it is.

Foxinio:
> BTW, this PR does two things: it adds Int64 type, and refactors the evaluator. Generally, I would prefer to have two separate PRs in such a case, but since the work is already done, you can leave it as it is.

I think since this refactor amounts to separating module into multiple files without really changing implementation, we can do two in one PR.

lib/Prelude.fram
```diff
@@ -123,3 +126,66 @@ pub let printInt   = extern dbl_printInt   : Int ->[IO] Unit
...
+  pub let zero = 0L
+  pub let one  = 1L
+  pub method succ {self : Int64} = self + one
+  pub method pred {self : Int64} = self - one
+
+  pub let ofInt (n : Int) = n.toInt64
+  pub method toString {self : Int64} =
+    (extern dbl_int64ToString : Int64 -> String) self
+  pub method toInt {`re : {type X} -> Unit ->[|_] X, self : Int64} =
+    if self <= 0x3fffffffffffffffL && self >= 0xc000000000000000L then
```
ppolesiuk, line 176: This range is still broken.

ppolesiuk (commented):
I would wait with merging this PR until #133 will be merged (or we decide that it's completely wrong). Then we will be able to cleanup a bit the Prelude. And there is a still one minor issue with this code (see the comment below).

lib/Prelude.fram
ppolesiuk: I see. It's correct, because `0xc000000000000000L` is a negative number, but it's very counterintuitive.

PR #117, opened 2024-06-03
Special syntax for nested method calls

### LINKED ISSUE #107: Method calls on the same precedence level as application
A syntactic sugar for chaining method calls would be nice. For instance
```
m.foo x y
  >.bar 42
  >.baz 13
```
would be syntactic sugar for `((m.foo x y).bar 42).baz 13`.

The choice of the syntax could be discussed. I like `>.` as in the above example, but other options I see are `|.` `|>.` `.>`. The `|.` is particularly interesting, because `|` has special meaning and the programmer cannot define such an operator.

Bohun9:
This pull request should resolve #107.

I am not sure what is the easiest way to get a position for `EMethod(e1,name)` in the `tr_expr` function, as names do not carry this information.


ppolesiuk (approved):
I approve these changes. I also fixed the issue mentioned in the description of this PR. Thanks for good piece of code.

PR #121, opened 2024-06-07
Basic language server implementation

pa000:
Provides a basic language server with a diagnostics reporting feature (shows you errors in the code).
Available with a new `framls` executable.

The server intercepts error reports made by the typechecker so the result should be the same as running the file with `dbl`.

The code structure makes it easy to add new capabilities (but they will require changes in the dbl code).

### DIFF
--- a/src/InterpLib/Error.ml
+++ b/src/InterpLib/Error.ml
@@ -17,7 +17,7 @@ let err_counter = ref 0
 let incr_error_counter () =
   err_counter := !err_counter + 1
 
-let report ?pos ~cls msg =
+let report_to_stderr ?pos ~cls msg =
   let name =
     match cls with
     | FatalError ->
@@ -35,6 +35,12 @@ let report ?pos ~cls msg =
   | Some pos ->
     Printf.eprintf "%s: %s: %s\n"
       (Position.to_string pos) name msg
+
+let report_impl = ref report_to_stderr
+
+let set_report_function f = report_impl := f
+
+let report ?pos ~cls msg = !report_impl ?pos ~cls msg
 
 let assert_no_error () =
   if !err_counter <> 0 then
--- a/src/Lsp/Connection.ml
+++ b/src/Lsp/Connection.ml
@@ -1,0 +1,59 @@
+(* This file is part of DBL, released under MIT license.
+ * See LICENSE for details.
+ *)
+
+(** Base HTML-like protocol *)
+
+open In_channel
+open Out_channel
+
+exception Connection_error of string
+
+type headers =
+  { content_length: int option 
+  ; content_type: string option
+  }
+
+let parse_header headers line =
+  match String.split_on_char ':' line |> List.map String.trim with
+  | ["Content-Length"; value] -> begin
+    match int_of_string_opt value with
+    | None -> raise (Connection_error ("Invalid Content-Length value: " ^ line))
+    | Some len -> { headers with content_length = Some len }
+    end
+  | ["Content-Type"; value] -> { headers with content_type = Some value }
+  | _ -> raise (Connection_error ("Invalid header: " ^ line))
+
+let rec collect_headers (ic: in_channel) =
+  match input_line ic with
+  | None -> raise (Connection_error "Unexpected end of file")
+  | Some "\r" -> []
+  | Some line -> line :: collect_headers ic
+
+let receive_headers (ic: in_channel) =
+  let headers = collect_headers ic in
+  List.fold_left parse_header
+    { content_length = None; content_type = None }
+    headers
+
+let receive_body (ic: in_channel) len =
+  match really_input_string ic len with
+  | None -> raise (Connection_error "Unexpected end of file")
+  | Some body -> body  
+
+let receive_string (ic: in_channel) =
+  let headers = receive_headers ic in
+  match headers.content_length with
+  | None -> raise (Connection_error "Missing Content-Length header")
+  | Some len -> receive_body ic len
+
+let output_line (oc: out_channel) message =
+  let line = message ^ "\r\n" in
+  output_string oc line
+
+let send_string (oc: out_channel) message_string =
+  let length = String.length message_string in
+  output_line oc ("Content-Length: " ^ string_of_int length);
+  output_line oc "";
+  output_string oc message_string; flush oc
+
--- a/src/Lsp/Connection.ml
+++ b/src/Lsp/Connection.ml
@@ -10,7 +10,7 @@ open Out_channel
 exception Connection_error of string
 
 type headers =
-  { content_length: int option 
+  { content_length: int option
   ; content_type: string option
   }
 
@@ -39,7 +39,7 @@ let receive_headers (ic: in_channel) =
 let receive_body (ic: in_channel) len =
   match really_input_string ic len with
   | None -> raise (Connection_error "Unexpected end of file")
-  | Some body -> body  
+  | Some body -> body
 
 let receive_string (ic: in_channel) =
   let headers = receive_headers ic in
--- a/src/Lsp/JsonRpc.ml
+++ b/src/Lsp/JsonRpc.ml
@@ -35,17 +35,18 @@ let rec run (state : State.t) handle_request handle_notification =
       send_response state response;
       loop state
     | Some { id; method_name; params; _ } ->
-      try match id with
-      | None ->
-        let notification = parse_notification method_name params in
-        let state = handle_notification state notification in
-        loop state
-      | Some id ->
-        let request = parse_request method_name params in
-        let state, result_or_error = handle_request state request in
-        let response = make_response ~id result_or_error in
-        send_response state response;
-        loop state
+      try
+        match id with
+        | None ->
+          let notification = parse_notification method_name params in
+          let state = handle_notification state notification in
+          loop state
+        | Some id ->
+          let request = parse_request method_name params in
+          let state, result_or_error = handle_request state request in
+          let response = make_response ~id result_or_error in
+          send_response state response;
+          loop state
       with
       (* Only send error response if the message was a Request *)
       | Type_error _ when Option.is_some id ->
--- a/src/Lsp/JsonRpc.ml
+++ b/src/Lsp/JsonRpc.ml
@@ -29,7 +29,7 @@ let rec run (state : State.t) handle_request handle_notification =
 
     let message_raw = receive_string ic in
     match parse_message_option message_raw with
-    | None -> 
+    | None ->
       let response = make_response
         (make_response_error ~code:ParseError ~message:"Parse error" ()) in
       send_response state response;
@@ -49,8 +49,8 @@ let rec run (state : State.t) handle_request handle_notification =
       with
       (* Only send error response if the message was a Request *)
       | Type_error _ when Option.is_some id ->
-        let error = make_response_error 
-            ~code:InvalidParams 
+        let error = make_response_error
+            ~code:InvalidParams
             ~message:"Invalid params" () in
         let response = make_response ?id error in
         send_response state response;
--- a/src/Lsp/JsonRpc.ml
+++ b/src/Lsp/JsonRpc.ml
@@ -1,0 +1,62 @@
+(* This file is part of DBL, released under MIT license.
+ * See LICENSE for details.
+ *)
+
+(** Main loop. Get message, handle it, send response. *)
+
+open Connection
+open Message
+
+let send_response state response =
+  let response_json = json_of_response response in
+  let response_string = Yojson.Safe.to_string response_json in
+  send_string (State.out_channel state) response_string
+
+let send_message state message =
+  let message_json = json_of_message message in
+  let message_string = Yojson.Safe.to_string message_json in
+  send_string (State.out_channel state) message_string
+
+let send_notification state notification =
+  let message = message_of_notification notification in
+  send_message state message
+
+let rec run (state : State.t) handle_request handle_notification =
+  let rec loop state =
+    let open Yojson.Safe.Util in
+
+    let ic = State.in_channel state in
+
+    let message_raw = receive_string ic in
+    match parse_message_option message_raw with
+    | None -> 
+      let response = make_response
+        (make_response_error ~code:ParseError ~message:"Parse error" ()) in
+      send_response state response;
+      loop state
+    | Some { id; method_name; params; _ } ->
+      try match id with
+      | None ->
+        let notification = parse_notification method_name params in
+        let state = handle_notification state notification in
+        loop state
+      | Some id ->
+        let request = parse_request method_name params in
+        let state, result_or_error = handle_request state request in
+        let response = make_response ~id result_or_error in
+        send_response state response;
+        loop state
+      with
+      (* Only send error response if the message was a Request *)
+      | Type_error _ when Option.is_some id ->
+        let error = make_response_error 
+            ~code:InvalidParams 
+            ~message:"Invalid params" () in
+        let response = make_response ?id error in
+        send_response state response;
+        loop state
+      (* The server MUST NOT reply to a Notification even in case of an error *)
+      | Type_error _ ->
+        loop state
+  in loop state
+
--- a/src/Lsp/State.ml
+++ b/src/Lsp/State.ml
@@ -3,7 +3,7 @@
  *)
 
 (** Server state *)
- 
+
 module UriMap = Map.Make(String)
 
 type uri = string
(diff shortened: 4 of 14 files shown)


ilikeheaps:
I've tested it shortly on my setup, seems to be working. Does highlight errors with description.

Steps I've taken to setup it:
1. I have updated the branch with master locally, will put up a branch in a moment so we're sure we work on the same version
2. Added yojson library and findlib to my development environment (Nix)
3. Made lsp client configuration for Emacs with lsp-mode, will post it somewhere

ilikeheaps:
Here is the updated branch: https://github.com/pa000/dbl/pull/1

ilikeheaps:
And for now I'll drop the Emacs configuration here:
`lsp.el`:
```
;; eval this file to setup framls for lsp-mode
(add-to-list 'lsp-language-id-configuration '(".*\\.fram$" . "fram"))

;; Emacs must be ran within environment with 'framls' available; either:
;; - install it (TODO 'dune install ???' -- I don't use it)
;; - or start Emacs with 'dune exec emacs'
(lsp-register-client (make-lsp-client
                      :new-connection (lsp-stdio-connection "framls")
                      :activation-fn (lsp-activate-on "fram")
                      :server-id 'framls))
```

ilikeheaps:
Oh right, CI will need to be updated with the new library.

src/InterpLib/Error.ml
```diff
@@ -56,6 +56,12 @@ let report ?pos ~cls msg =
   | None, _ ->
     Printf.eprintf "%s: %s\n" name msg
 
+let report_impl = ref report_to_stderr
```
ilikeheaps, line 59: I'm not a fan of mutable state. Do we need to dynamically change this? Perhaps I'll answer this myself as I read further.

src/Lsp/Connection.ml
```diff
@@ -0,0 +1,59 @@
...
+open Out_channel
+
+exception Connection_error of string
+
+type headers =
+  { content_length: int option
+  ; content_type: string option
+  }
+
+let parse_header headers line =
```
ilikeheaps, line 17: This sounds like something we could use a third party library for unless it's pretty custom. (could refactor later)

src/Lsp/Uri.ml
```diff
@@ -0,0 +1,9 @@
+(* This file is part of DBL, released under MIT license.
+ * See LICENSE for details.
+ *)
+
+(** URI helper functions *)
+
+let to_path uri =
+  String.sub uri (String.length "file://")
+    (String.length uri - String.length "file://")
```
ilikeheaps, line 9: Perhaps it should check that the prefix is indeed `"file://"`?

ilikeheaps (commented):
Checked some files, so far so good (left couple ideas to improve type safety, could be dismissed or done later).

src/InterpLib/Error.ml
pa000: I don't think anyone here is. However, without any mutable state we would have to pass a `report` argument to every type-checker function which seems impractical. Regular DBL code has no reason to change this variable and the language server only does it before running the type-checker, so I think there is little risk with using a ref here.

src/Lsp/Connection.ml
pa000: I'm not a fan of dependencies and the whole file is like 60 lines (and does exactly what's needed), so I'm not sure how much a library would help.

src/Lsp/Uri.ml
pa000: I'll provide an update which uses the `uri` library to manage URIs (the current version doesn't work with percent-encoded strings), so this file will be gone. That said, I'm not sure what to do if someone provides a URI with a scheme different than `file` (e.g `http`). I don't think it's worth handling such cases.

pa000: It seems that the only place where the path to the file is actually needed is when setting the directory to look for modules, so in the worst case it will just not find other modules. I think I'll leave it like that.

pa000:
I believe I'm finished with the final clean-ups (that seem to have made the PR 500 lines bigger oops) and the code is now ready to be thoroughly reviewed.

src/Lsp/framls.ml
```diff
@@ -0,0 +1,110 @@
...
+
+(** Type-check given file and send the diagnostics to the client *)
+let process_program state uri =
+  let real_path   = Uri.path uri                      in
+  let temp_path   = State.get_document_path state uri in
+  let diagnostics = ref []                            in
+
+  InterpLib.Error.set_report_function (report temp_path real_path diagnostics);
+  set_module_dirs real_path;
+  type_check temp_path;
```
ilikeheaps, line 70: Why do we typecheck the temporary file instead of the real one? Does temporary one include changes from the editor that are not saved yet?

ilikeheaps (commented):
Will probably need one more pass to finish the review. Nothing major in this batch of comments, mostly "I think this could improve readability although I'm not sure and it's not a big deal".

src/Lsp/Connection.ml
```diff
@@ -0,0 +1,69 @@
...
+ *)
+
+(** Base HTTP-like protocol *)
+
+(** The JSON-RPC messages are sent over this simple protocol.
+  It consists of a header and a content part, separated by "\r\n":
+    header1: value1\r\n
+    header2: value2\r\n
+    \r\n
+    [JSON-RPC message]
```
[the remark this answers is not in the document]
pa000: It's an implementation detail (not stated in the specification) and the top comment wasn't meant for that.

src/Lsp/Connection.ml
```diff
@@ -0,0 +1,69 @@
...
+  match String.split_on_char ':' line |> List.map String.trim with
+  | ["Content-Length"; value] -> begin
+    match int_of_string_opt value with
+    | None -> raise (Connection_error ("Invalid Content-Length value: " ^ line))
+    | Some len -> { headers with content_length = Some len }
+    end
+  | ["Content-Type"; value] -> { headers with content_type = Some value }
+  | _ -> headers
+
+let rec collect_headers (ic: in_channel) =
```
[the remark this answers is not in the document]
pa000: Yes, `collect_header_lines` is better.

src/Lsp/README.md
```diff
@@ -0,0 +1,136 @@
+framls
+======
+
+`framls` is a language server for Fram. This document is meant to be
+an entry point for someone wanting to contribute to `framls`.
+If you want to configure `framls` for use with your code editor please visit
+[fram-lang.org](https://fram-lang.org).
```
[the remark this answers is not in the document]
pa000: No because we don't have any editor extensions yet.

src/Lsp/README.md
```diff
@@ -0,0 +1,136 @@
...
+Base is a simple HTTP-like protocol. It consists of a header and content part,
+separated by "\r\n":
+```
+header1: value1\r\n
+header2: value2\r\n
+\r\n
+[JSON-RPC message]
+```
+
+It is implemented in the `Connection` module.
```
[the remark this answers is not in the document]
pa000: I think we will because it probably won't change. I think it's useful to have it both in README (to get an overview) and in each file on its own (so you don't have to jump between files just to read the doc comment).

src/Lsp/JsonRpc.ml
```diff
@@ -0,0 +1,73 @@
...
+
+let send_response state response =
+  let response_json = json_of_response response in
+  let response_string = Yojson.Safe.to_string response_json in
+  send_string (State.out_channel state) response_string
+
+let send_message state message =
+  let message_json = json_of_message message in
+  let message_string = Yojson.Safe.to_string message_json in
+  send_string (State.out_channel state) message_string
```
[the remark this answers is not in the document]
pa000: sure, now it's very dry.

src/Lsp/JsonRpc.ml
```diff
@@ -0,0 +1,73 @@
...
+(* ------------------------------------------------------------------------- *)
+(* Main function *)
+
+let rec run state handle_request handle_notification =
+
+  (* Handle messages in a loop *)
+  let rec loop state =
+    let string = receive_string (State.in_channel state) in
+    match message_of_string string with
+    (* There was an error parsing the message *)
```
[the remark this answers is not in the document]
pa000: It's definitely a parsing error (the string is already received). I don't think a custom type is a good idea because the error is produced by the callee and we're just passing it over, so we don't care what kind of error it is.

src/Lsp/State.ml
```diff
@@ -0,0 +1,64 @@
...
+    output_string oc content
+  );
+  let document = { temp_path } in
+  { state with documents = UriMap.add uri document state.documents }
+
+let update_document state uri new_content =
+  match UriMap.find_opt uri state.documents with
+  | None -> state
+  | Some doc ->
+    Out_channel.with_open_text doc.temp_path (fun oc ->
```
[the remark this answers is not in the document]
pa000: that's the expected semantics of opening a file but sure.

src/Lsp/framls.ml
pa000: Yes, exactly. I will add that to the doc comment of `State.get_document_path`. Do you think it should be explained anywhere else? It's important to know the difference between the temp and real files.

src/Lsp/Connection.ml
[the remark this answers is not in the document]
ilikeheaps: Fair enough.

src/Lsp/State.ml
```diff
@@ -0,0 +1,65 @@
+(* This file is part of DBL, released under MIT license.
+ * See LICENSE for details.
+ *)
+
+(** Server state *)
+
+module UriMap = Map.Make(Uri)
+
+(** We keep the latest version of each document in a temp file *)
```
ilikeheaps, line 9: I think it's worth clarifying what "latest" means. Not sure if my suggestion is clear enough, natural language processing part of my brain is shutting down now. Also alternatively could have some dictionary of "domain specific" phrases.
```suggestion
(** We keep the latest version (being edited, possibly unsaved) of each document in a temp file *)
```
or
```suggestion
(** We keep the latest version (possibly dirty) of each document in a temp file *)
```

src/Lsp/State.ml
```diff
@@ -0,0 +1,65 @@
...
+  | None -> state
+  | Some doc ->
+    (* overwrites the file if it exists *)
+    Out_channel.with_open_text doc.temp_path (fun oc ->
+      output_string oc new_content
+    ); state
+
+let close_document state uri =
+  match UriMap.find_opt uri state.documents with
+  | None -> state
```
ilikeheaps, line 52: I wonder if we shouldn't do something more drastic in this case, maybe even fail catastrophically. What could lead to this situation? (question applies to surrounding functions too)

src/Lsp/State.ml
ilikeheaps: Or is it obvious from description in `State.mli`? (I haven't got the habit to read .mli and .ml in parallel)

src/Lsp/State.mli
```diff
@@ -0,0 +1,38 @@
+(* This file is part of DBL, released under MIT license.
+ * See LICENSE for details.
+ *)
+
+(** Server state *)
+
+(** The LSP client sends a notification when the user opens, updates
```
ilikeheaps, line 7: What is an "update" from the point of view of an editor? Is there some standard or convention or is it arbitrary? I suppose I could imagine a range of behaviours:
1. send update on every character input
2. send update on every word completed
3. every sentence completed
4. when the file is saved

src/Lsp/framls.ml
ilikeheaps: I was thinking it could be sufficient to document that in `type document` definition but now I realized this code avoids using `document` type. Huh.

I don't think any other place produced the same question though.

src/Lsp/Message.mli
```diff
@@ -0,0 +1,224 @@
...
+
+type position = {
+  line      : int;
+  character : int;
+}
+
+type range = {
+  start : position;
+  end_  : position (* json key: end *);
+}
```
ilikeheaps, line 23: Is this inclusive?

src/Lsp/Message.mli
```diff
@@ -0,0 +1,224 @@
...
+  in the LSP specification, together with needed conversion functions
+  and smart constructors. *)
+
+type json = Yojson.Safe.t
+
+type uri = Uri.t
+
+type position = {
+  line      : int;
+  character : int;
```
ilikeheaps, line 17: Not sure if `character` is obvious enough without the context of contrasting `line`... Is this from LSP specification?

src/Lsp/Message.mli
ilikeheaps, line 1: Code in these (.ml and .mli) seems straightforward but I think it could use more documentation, I'm not sure what some types/fields are supposed to be. I suggest either copying from LSP documentation or linking the relevant parts.

src/Lsp/framls.ml
```diff
@@ -0,0 +1,110 @@
...
+          "textDocumentSync": {
+            "openClose": true,
+            "change": 1
+          }
+        },
+        "serverInfo": {
+          "name": "framls"
+        }
+      }
+    |}
```
ilikeheaps, line 27: Could this use JSON constructors from `Message.ml`?

ilikeheaps (approved):
Okay, I've seen it all. I left some more minor questions/suggestions but it looks good enough to me.

src/Lsp/State.ml
pa000: It's indeed explained in `State.mli` (or at least I tried). If someone wants to edit the `.ml` file they should know what latest means. Because of that I also removed this comment.

src/Lsp/State.ml
```diff
@@ -0,0 +1,65 @@
...
+let create ~in_channel ~out_channel =
+  {
+    connection = { in_channel; out_channel };
+    documents = UriMap.empty;
+  }
+
+let out_channel { connection = { out_channel; _ }; _ } = out_channel
+let in_channel { connection = { in_channel; _ }; _ } = in_channel
+
+let open_document state uri content =
```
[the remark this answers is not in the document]
pa000: You're right.

src/Lsp/State.ml
pa000: This case shouldn't happen. If it does that probably means an error happened on our side and so we should fail, you're right.

src/Lsp/State.mli
pa000: It's up to the client but most probably 1 (anything else would likely make the lsp feel laggy).

src/Lsp/Message.mli
pa000: [start, end)
https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#range

src/Lsp/Message.mli
pa000: [Yes](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#position)

src/Lsp/Message.mli
pa000: I'd link the specification for each type but the url is crazy long (e.g. https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#range). If someone wants to implement an LSP feature they should probably have the specification open regardless.

src/Lsp/framls.ml
pa000: No (but it could be added). This message is a bit different because it depends on the capabilities that the server supports. It'd be best to create this message in a way that avoids inconsistencies (e.g. someone extends the server with some capability but forgets to update it here). I added a comment explaining it. It's fine for now.

ilikeheaps:
Whoop, seems like I can't mark conversations as resolved, only single comments of my own.

pa000:
That's all from me. We can finally merge it.

src/Lsp/JsonRpc.ml
```diff
@@ -0,0 +1,73 @@
...
+
+let send_error state ?id code message =
+  let error = make_response_error ~code ~message () |> Result.error in
+  let response = make_response ?id ~result_or_error:error () in
+  send_response state response
+
+(* ------------------------------------------------------------------------- *)
+(* Main function *)
+
+let rec run state handle_request handle_notification =
```
ppolesiuk, line 36: This function is not recursive (but uses recursive `loop` function). The `rec` keyword is not necessary here.

ppolesiuk (commented):
The LSP looks good, and I would like to merge it. However, two minor issues should be resolved before merging:
- code formatting is not fully consistent with the existing code base (I can fix it just before the merge),
- the LSP adds some dependencies to the project. It would be nice (if possible) to make them optional, i.e., it the user doesn't have required libraries, he still can build and install DBL, but `framls` executable will not be available.

pa000:
1. You can fix the formatting.
2. `framls` is now optional. To do that I needed to change the default build target in dune to `install`. I don't know what that means but it works.

Issue #11, opened 2024-03-23
Reisntatniation of named parameters in implicit instantiation resolution


ppolesiuk:
Currently, regular named parameters can be implicitly introduced, but not instantiated. This prevents an instantiation of an implicit parameter of type `{ x : A} -> B`, assuming that there is such an implicit in the scope. We could allow the instantiation algorithm to instantiate named parameters that were **recently** introduced by `Env.open_scheme`.

PR #122, opened 2024-06-11
Optional params

### LINKED ISSUE #28: Optional parameters

Tagada14:
Introduces optional parameters.

The implementation is finished and appears to be working.

~~Things left to do~~:
- [x] Add missing comments
- [x] Add tests
- [x] Check for Col limit
- [x] Resolve conflicts
- [x] ~~Split fixes and code clean-up of code not related to this PR into a separate PR~~
- [x] Potentially rewrite parts of the code to make it more readable
- [x] Rework the PreludeTypes to check if a correct version of Option type is in the environment

Nice to have:
- [x] Create a test case for errors

resolves #28


Tagada14:
@ppolesiuk The PR is ready for review

src/TypeInference/ExprUtils.ml
```diff
@@ -155,6 +155,11 @@ and instantiate_named_param ~nset ~inst env (e : T.expr) (name, isch) =
   | None, T.NVar x ->
     (* TODO: we could provide freshly bound parameters here *)
     Error.fatal (Error.unbound_named_param ~pos:e.pos x)
+  | None, T.NOptionalVar x ->
+    assert (T.Scheme.is_monomorphic isch);
+    let arg = PreludeTypes.mk_None ~env ~pos:e.pos isch.sch_body in
+    { T.pos = e.pos; T.data = T.EApp(e, arg) }
```
ppolesiuk, line 161: This implementation suffers problem similar to #11: implicit arguments are not reinstatiated when function has polymorphic argument. Here is an example code that exhibit this behavior:
```
let foo (f : {?x:Int} -> Unit ->[] Int) = f {x=42} ()
let bar {?x} =
  match x with
  | None => 0
  | Some y => y
  end

let z = foo bar // z is equal to 0, but intuitively it should give 42
```
We are not going to fix it now. We will fix it together with #11.

src/TypeInference/Pattern.ml
```diff
@@ -330,6 +331,26 @@ and check_pattern_schemes ~env ~scope pats schs =
...
+  let (env, pat, _, r_eff) =
+    check_scheme ~env ~scope pat sch in
+  (env, pat, sch, r_eff)
+| ArgPattern pat ->
+  let tp = Env.fresh_uvar env T.Kind.k_type in
+  let tp = (PreludeTypes.mk_Option ~env ~pos:pat.pos tp) in
+  let scope = Env.scope env in
+  let (env, pat, _, r_eff) = check_type ~env ~scope pat tp in
+  (env, pat, T.Scheme.of_type tp, r_eff)
+
```
ppolesiuk, line 353: Indentation of this piece of code should be increased

src/TypeInference/Pattern.ml
```diff
@@ -330,6 +331,26 @@ and check_pattern_schemes ~env ~scope pats schs =
 
   | [], _ :: _ | _ :: _, [] -> assert false
 
+let infer_optional_arg_scheme env (arg : S.arg) =
+match arg with
+| ArgAnnot(pat, sch) ->
+  let sch_pos = sch.sch_pos in
+  let sch = Type.tr_scheme env sch in
+  assert (T.Scheme.is_monomorphic sch);
```
ppolesiuk, line 339: There is no invariant that ensure this. User can write an invalid program that violates this assertion:
```
let foo {?x:{x:Int} -> Int} = x
```

The implementation should check if this scheme is monomorphic, and if not, display a readable error for the user.

src/TypeInference/Pattern.mli
```diff
@@ -23,6 +23,12 @@ val check_type :
 val infer_arg_scheme :
   Env.t -> S.arg -> Env.t * T.pattern * T.scheme * ret_effect
 
+(** Infer type-scheme of given formal optional argument. Returns extended
+  environment, a pattern that represents an argument, its scheme, and the
+  effect of pattern-matching *)
+val infer_optional_arg_scheme :
+  Env.t -> S.arg -> Env.t * T.pattern * T.scheme * ret_effect
```
ppolesiuk, line 30: Should this function be public? I didn't find any place where it was used outside this module.

src/TypeInference/PolyExpr.ml
```diff
@@ -137,25 +137,45 @@ let rec check_explicit_insts ~tcfix
...
+            end
+          | _ -> None
+          end 
+        in
+        match res with
+        | Some(n, e, tp, r_eff1) -> (n, e, T.Scheme.of_type tp, r_eff1)
+        | None -> 
+          Error.warn (Error.redundant_named_parameter ~pos n);
+          let (e, tp, r_eff1) = infer_expr_type env e eff in
+          (n, e, T.Scheme.of_type tp, r_eff1)
```
ppolesiuk, line 178: The logic behind this part of code has grown so it is hard to follow. Maybe it could by improved by a simple comment explaining what's going on.

src/TypeInference/PreludeTypes.ml
```diff
@@ -0,0 +1,72 @@
...
+let mk_Some ~env ~pos tp_arg expr_arg : T.expr =
+  let make data pos = { T.data; T.pos } in
+  let tp' = mk_Option ~env ~pos tp_arg in
+  let _, x = option_to_whnf ~env ~pos tp' in
+  match Env.lookup_adt env x with
+  | Some adt_info -> (
+      let { Module.adt_proof; adt_args; adt_ctors; adt_type } = adt_info in
+      match
+        List.find_index
+          (fun (ctor_decl : T.ctor_decl) -> String.compare ctor_decl.ctor_name "Some" == 0)
```
ppolesiuk, line 41: It's easier to use standard equality to compare strings: `ctor_decl.ctor_name = "Some"`

src/TypeInference/PreludeTypes.ml
```diff
@@ -0,0 +1,72 @@
...
+          adt_ctors
+      with
+      | Some idx ->
+          make
+            (T.ECtor
+               ( make (T.ETApp (adt_proof, tp_arg)) Position.nowhere,
+                 idx,
+                 [],
+                 [ expr_arg ] ))
```
ppolesiuk, line 50: Do we know, that `Some` constructor expects one parameter? Should it be checked? I know, such errors should never happen for casual user, but ...

src/TypeInference/TypeUtils.ml
```diff
@@ -48,7 +48,8 @@ let open_scheme ~pos (env : Env.t) (sch : T.scheme) =
         let (env, x) =
           match name with
           | T.NLabel -> Env.add_the_label_sch env sch
-          | T.NVar x -> Env.add_poly_var env x sch
+          | T.NVar x  -> Env.add_poly_var env x sch
+          | T.NOptionalVar x -> Env.add_mono_var env x sch.sch_body
```
ppolesiuk, line 52: `add_mono_var env x sch.sch_body` and `add_poly_var env x sch` are equivalent for monomorphic types (for polymorphic types the former doesn't have any sense). Maybe these two clauses could be merged?

test/err/tc_0007_missingOptionTypeDef.fram
ppolesiuk, line 1: In negative tests, we try to write some comments with regular expressions that should be matched with the stderr output (see other tests).

test/ok/ok0109_optionalParams.fram
```diff
@@ -0,0 +1,23 @@
...
+
+let f {?x} = x
+let a = f
+let b = f {x=2}
+
+let g {?x : Int} () = 
+match x with
+| Some n => n + 2
+| None => 42
+end
```
ppolesiuk, line 13: We should format tests correctly. This match expression should be indented. The next (in lines 18-21) should be fixed too.

ppolesiuk (changes requested):
The implementation looks very solid. Thanks for good job! There are some minor things that should be addressed. Then, I will merge these changes into upstream.

src/TypeInference/PreludeTypes.ml
ppolesiuk: I tried to implement it, checking correct type of `Some` and `None` constructor. The code is too complicated for a thing that should never happen, so I decided to not address this issue. Internal type error is good enough.

PR #123, opened 2024-06-12
String interpolation

wojpok:
This is my take on string interpolation.

This change requires additional state in lexer, as discussed already. Most of the work is made while "desugaarring".
 
Since optional parameters are still not implemented, this is a draft pull request and will be updated later. 
For demonstration purposes I added a new ```toFormatted``` method, that is invoked while interpolating strings.


ppolesiuk (changes requested):
Thanks! The implementation looks good, but has several flaws.
- Lexer is buggy. See comments below.
- Changes to standard library are too complicated for me.
- Code formatting could be improved. There are lines longer than 80 columns.
- I'm not sure if the implementation works well in REPL. I didn't find a call to `Lexer.reset` in REPL related functions.
- Position tracking in `Desugar` could be improved.

BTW, I think we can rename `toFormatted` method to simpler `format`. Do you agree?

Since this PR was open for a long time, I can fix the issues mentioned in this review, and adapt it to the new codebase.

wojpok:
I have resolved most of the issues. We definitely should discuss how to approach formatting.  I have decided to use `toString` method if no format argument is provided and `format` method otherwise. I think more elaborate formatting could be implemented alongside pretty printing library

code.rkt
ppolesiuk, line 1: This file looks like dump from some debugging. It should not be present in the repository.

src/Eval/External.ml
```diff
@@ -110,7 +120,7 @@ let extern_map =
     "dbl_strLen",  str_fun (fun s -> VNum (String.length s));
     "dbl_strGet",  str_fun (fun s -> int_fun (fun n -> VNum (Char.code s.[n])));
     "dbl_chrToString",  int_fun (fun c -> VStr (Char.escaped (Char.chr c)));
-    "dbl_chrListToStr", list_chr_fun (fun xs -> VStr (List.to_seq xs |> String.of_seq));
+    "dbl_chrListToStr", list_fun to_char (fun xs -> VStr (List.to_seq xs |> String.of_seq));
```
ppolesiuk, line 123: This line is too long

src/DblParser/Desugar.ml
```diff
@@ -526,8 +530,30 @@ and tr_expr (e : Raw.expr) =
...
+    let tr_format (expr : Raw.expr) (fmt : Raw.expr option) = 
+      let format = match fmt with
+      | None   -> 
+        { pos = expr.pos; data = Raw.ECtor "None" }
+      | Some fmt ->  
+        { pos = expr.pos
+        ; data = Raw.EApp (with_nowhere (Raw.ECtor "Some"), [fmt]) }
+        in
+      let fld = make (Raw.FldNameVal (Raw.NOptionalVar "fmt", format)) in
+      let mth = make (Raw.EMethod (expr, "format")) in
```
ppolesiuk, line 543: Using `make (Raw.EMethod ...)` gives us imprecise information about position. For instance:
```
> data T = A ;;
> "\{ A }" ;;
fatal error: Type T has no method named format
   | "\{ A }" ;;
   | ^^^^^^^^
   |
>
```
It would be better to point to `expr` instead of the whole interpolation.

src/DblParser/Desugar.ml
```diff
@@ -526,8 +530,30 @@ and tr_expr (e : Raw.expr) =
...
+  | EInterp (s, xs) ->
+    let tr_format (expr : Raw.expr) (fmt : Raw.expr option) = 
+      let format = match fmt with
+      | None   -> 
+        { pos = expr.pos; data = Raw.ECtor "None" }
+      | Some fmt ->  
+        { pos = expr.pos
+        ; data = Raw.EApp (with_nowhere (Raw.ECtor "Some"), [fmt]) }
+        in
+      let fld = make (Raw.FldNameVal (Raw.NOptionalVar "fmt", format)) in
```
ppolesiuk, line 542: It think the point of using optional arguments here is to omit them when they are not provided instead of explicitly provide `None`.

ppolesiuk (changes requested):
There are still some minor issues with desugaring. For other parts of this PR, I think they are ready for merge.

wojpok:
Issues have been fixed. I also noticed that formatting argument had have wrong position assigned in case of type error. I fixed it by attaching argument's position to `Some` constructor

ppolesiuk (approved):
Now the PR is ready to merge. Thanks!

In the last commit I slightly improved position tracking and changed `fmt` parameter to regular (`fmt = e` instead of `?fmt = Some e`). For optional parameters it means the same, but it allows `format` methods with mandatory `fmt`. Moreover, the code is simpler.

PR #124, opened 2024-06-13
Fix visibility of named patterns

forell:
Public visibility was not being properly propagated to named parameters in patterns. For example, the following code would fail.
```
data T = C of { S, x : Int }

module M
  pub let C { S, x } = C { S = Unit, x = 42 }
end

let y = M.x

let f (x : M.S) = x
```
Thanks @ppolesiuk for pointing out this problem.


ppolesiuk (approved):
Looks reasonable. Thanks!

PR #125, opened 2024-06-13
Fix `!=` operator in lexer

### LINKED ISSUE #108: Implement `!=` operator
The `!=` operator is defined in https://github.com/fram-lang/dbl/blob/c17f2fb726c96c59daf25cbfc874f9fe36815eb8/lib/Prelude.dbl#L27 but it is not implemented in the lexer and parser yet.

ppolesiuk:
- `OP_230` and `expr_230` were removed
- Operators starting with `!` are treated as binary at level 50.

Rationale: The `OP_230` was introduced to have OCaml-like dereference operators (`!x`) on a very high precedence level. However, since we have methods in Fram, we can write `x.get`, which has high precedence, so unary dereference operator is not needed.

Fix #108


forell (approved):
Looks good to me. I won't be missing the unary `!`, I think `x.get` is short enough for a language which doesn't encourage mutability as the first choice and the precedence works out well.

PR #135, opened 2024-06-18
Better printing of error position in REPL and with colors

### LINKED ISSUE #134: Better position printing in REPL
In #113 we improved position printing in batch mode. We should do the same in REPL. This require adding some buffer that remembers the last REPL command, and maybe some refactor of the `Position` module. The problem that I see is how to print a position that refers to some of previous commands. Currently, it is not perfectly correct, for instance:
```
> data T = A ;;
> data T = B of T | C ;;
> B C ;;
<stdin>:1:3: error: This expression has type T, but an expression was expected of type #T
  Type #T is defined at <stdin>:1:1-10
>
```

Foxinio:
It's a work in progress with emphasis on "in progress". Im sure it requires a lot of refactoring, because this changes brought a big mess into the codebase, but it works.

Fixes #134

Example:
```
$ dbl                
> false ;;
fatal error: Unbound variable false
   | false ;;
   | ^^^^^
   | 
 ```

### CODE REVIEW

**Comment by Foxinio** [comment]
This is work in progress, some bugs have been found in this implementation and are being fixed

**Comment by Foxinio** [comment]
> This is work in progress, some bugs have been found in this implementation and are being fixed

File: src/dbl.ml [hunk 1]
```diff
@@ -46,6 +46,10 @@ let cmd_args_options = Arg.align
     "-I",
     Arg.String (fun p -> cli_local_search_dirs := p :: !cli_local_search_dirs),
     " Add a path to local search directories";
+
+    "-color",
+    Arg.String (fun s -> DblConfig.print_colors_of_string s),
```
ppolesiuk, line 51: `Arg.spec` has `Symbol` constructor. It would fit better here.

src/TypeInference/PolyExpr.ml
```diff
@@ -39,7 +39,7 @@ let method_call_ctx ~pos ~env ~self ~self_tp ~self_r_eff ~eff =
     (result_expr, res_tp, ret_effect_join self_r_eff r_eff)
   | _ ->
     (* Method must be an arrow with monomorphic argument *)
-    InterpLib.InternalError.report ~reason:"invalid method type" ()
+    Error.fatal (Error.invalid_method_type ~pos)
```
ppolesiuk, line 42: Is it error or internal error? Do you have any example input that causes this error?

src/InterpLib/TextRangePrinting.ml
```diff
@@ -0,0 +1,219 @@
...
+    ; "effrow"; "else"; "end"; "extern"
+    ; "finally"; "fn"; "handle"; "handler"
+    ; "if"; "implicit"; "import"; "in"
+    ; "label"; "let"; "match"; "method"
+    ; "module"; "of"; "open"; "pub"
+    ; "rec"; "return"; "then"; "type"
+    ; "with"; "_"; "}"; "{"
+    ; "->"; "::"; "&&"; "||"; ";;"
+    ; "\\["; "\\]"; "=>"; "="; "|"
+    ] |> String.concat "\\|" in
```
ppolesiuk, line 104: This code seems to be hard to maintain. Can we extract keywords from the lexer?
I see that there would be a small technical difficulty related to cyclic dependency of modules.

ppolesiuk (changes requested):
The implementation looks good. However a few things should be fixed or at least discussed.

src/DblParser/Main.ml
```diff
@@ -23,7 +23,15 @@ let rec repl_seq imported () =
 and repl_seq_main imported () =
   flush stderr;
   Printf.printf "> %!";
-  let lexbuf = Lexing.from_channel stdin in
+  let buffer = Buffer.create 512 in
+  let bytes = Bytes.create 512 in
+  let fn buf n =
+    let res = input stdin bytes 0 n in
```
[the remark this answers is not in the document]
Foxinio: Before we were using Lexing.from_channel, and internally it is implemented using from_function, and line 29 is taken from there 1:1. 512 comes from there as well, and it is guaranteed there that n is always `Bytes.length bytes`.
But using `buf` is not a bad idea.

src/DblParser/Main.ml
```diff
@@ -35,14 +43,20 @@ and repl_seq_main imported () =
 
   | Raw.REPL_Expr e ->
     let def = make_nowhere (Lang.Surface.DReplExpr(Desugar.tr_expr e)) in
+    let str = Bytes.to_string @@ Buffer.to_bytes buffer in
+    InterpLib.Error.repl_input := str;
```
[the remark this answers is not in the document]
Foxinio: That doesn't seem to be the case since new buffer is created on every input, but its true that it can be created once and cleared before taking in any input.

src/InterpLib/TextRangePrinting.ml
Foxinio: I don't see a way around the dependency cycle, however I don't know that it will be so hard to maintain. I don't suppose we will be adding keywords that often and so many, so minor change once in a while won't be a problem.
When this list becomes outdated consequences will be apparent and harmless at the same time, also extremely easy to fix.

At the same time this isn't that necessary, there won't be that much loss in removing this part.

src/TypeInference/PolyExpr.ml
Foxinio: This change might have been done without proper understanding

src/InterpLib/TextRangePrinting.ml
Foxinio: I'm leaving this decision to you.

src/DblParser/Main.ml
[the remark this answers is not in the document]
ppolesiuk: That's true. At first, I didn't understand the code correctly. A new buffer is created for each input. I prefer the previous version, with local buffers instead of the global one.

src/InterpLib/TextRangePrinting.ml
ppolesiuk: I decided to remove this part. There are at least three reasons of this decision:

1. This code would be hard to maintain
2. It's far from being perfect: e.g., it colors keywords inside string literals
3. It provides functionality that is nice, but not necessary

ppolesiuk (approved):
I approve these changes. I will make final fixes before merging into upstream.

ppolesiuk:
I've made final fixes and resolves merge conflicts. In a last moment before actual merge, I've found another bug. Lexer errors are not printed correctly:
```
> 2 + 2 ;;
: Int
= 4
> ą ;;
fatal error: Invalid character `\196'
   | 2 + 2 ;;
   | ^
   | 
> 
```
I propose to fix it, by changing the type of `InterpLib.Error.repl_input` from `string ref` to `Buffer.t` and use it in `repl_seq_main` instead of `repl_buffer`. Does it sound reasonable?

Foxinio:
looks good to me

PR #139, opened 2024-07-02
Os stdlib module and separation of Eval/External module

Foxinio:
Looking at external function list made me think that there are too many
of them for one file, so i seperated them into modules making it easier
for extending functionality.

Implemented functionality that will allow for some system manipulation,
accessing environment variables, program arguments and system time.
Can be extended if there will be need for it.

This PR is dependent on #126


lib/Os.fram
```diff
@@ -0,0 +1,87 @@
+import open Mutable
+import List
+
+pub let argv = (extern dbl_argv : Unit -> ConstArray String) ()
+
+pub let getEnv = (extern dbl_getEnv : String -> Option String)
+
+pub let fileExists = (extern dbl_fileExists : String ->[IO] Bool)
+
+pub let isDirectory {`re : {X} -> Unit -> [|_]X } name =
```
Florian3k, line 10: To be consistent with existing code:
```suggestion
pub let isDirectory {`re : {X} -> Unit ->[|_] X } name =
```

lib/Os.fram
```diff
@@ -0,0 +1,87 @@
...
+  if fileExists name
+  then (extern dbl_isDirectory : String ->[IO] Bool) name
+  else `re ()
+
+pub let isRegularFile {`re : {X} -> Unit -> [|_]X } name =
```
Florian3k, line 15: Same here:
```suggestion
pub let isRegularFile {`re : {X} -> Unit ->[|_] X } name =
```

lib/Os.fram
```diff
@@ -0,0 +1,87 @@
...
+  , day : Int
+  , hour : Int
+  , minute : Int
+  , second : Int
+  }
+
+pub let getUnixTime = (extern dbl_getUnixTime : Unit ->[IO] Int)
+
+(* calculate date and time stamp from unix time *)
+pub let getTime {`re : {X} -> Unit -> [|_]X } () =
```
Florian3k, line 48: and here:
```suggestion
pub let getTime {`re : {X} -> Unit ->[|_] X } () =
```

dominik-muc (changes requested):
`Eval.ml` in `src` can be deleted.

Rather than prefix all external helper modules with External, I would just make a new directory, maybe rename Helpers to Utils?

`Os.fram` lacks rw operations on files, but that can be added in another PR.

Foxinio:
> Rather than prefix all external helper modules with External, I would just make a new directory, maybe rename Helpers to Utils?

The problem with making `External` a seperate module is that all externals depend on `Value.ml` and it could work, but would make this hell, so in my opinion it's fine as it is.

dominik-muc (approved):

wojpok (approved):

forell (changes requested):
I haven't really looked at the extern function refactor, but I left some comments about the `Os` module.
One last formatting nitpick: some of the files have extraneous newlines at the end.

lib/Os.fram
```diff
@@ -0,0 +1,116 @@
...
+  , hour : Int
+  , minute : Int
+  , second : Int
+  }
+
+{## Get the current time in Unix format. ##}
+pub let getUnixTime = (extern dbl_getUnixTime : Unit ->[IO] Int)
+
+{## Get the current time in Time type. ##}
+pub let getTime () =
```
forell, line 74: For some reason this function starts counting months at `-1` (maybe). Today's month is `3`, while the month for the unix time `1737632040`, which was in January, is `-1`. It would also help to specify that the returned time is in UTC (presumably).

lib/Os.fram
```diff
@@ -0,0 +1,60 @@
+{# This file is part of DBL, released under MIT license.
+ # See LICENSE for details.
+ #}
+
+import open Mutable
```
forell, line 5: Probably unused now.

PR #140, opened 2024-07-04
Fix visibility of module patterns

### LINKED ISSUE #138: Visibility does not affect module patterns
In the following example, module `N` should be public in module `M`, but isn't.
```
module M
  data T = C of { x : Int }
  pub let C { module N } = C { x = 42 }
end

let x = M.N.x
```

forell:
Fixes #138.


ppolesiuk (approved):
This fix looks good. I approve these changes.

PR #141, opened 2024-08-07
Return and finally clauses in first class handlers

ppolesiuk:
Return and finally clauses now are attached to first class handlers, making possible to define more intuitive first class handlers for e.g. state or backtracking. The delimiter's type and effect are not visible in the type of first-class-handlers, therefore, the dynamic labels of handlers do not work.


src/TypeInference/Unification.ml
```diff
@@ -276,14 +280,20 @@ let rec check_subtype env tp1 tp2 =
...
+    check_subeffect env oeff1 oeff2;
     let (env, a) = Env.add_anon_tvar env T.Kind.k_effect in
     let sub1 = T.Subst.rename_to_fresh T.Subst.empty a1 a in
     let sub2 = T.Subst.rename_to_fresh T.Subst.empty a2 a in
     check_subtype env (T.Type.subst sub1 tp1) (T.Type.subst sub2 tp2);
-    unify env (T.Type.subst sub1 rtp1) (T.Type.subst sub2 rtp2);
-    unify_at_kind env
-      (T.Type.subst sub1 reff1) (T.Type.subst sub2 reff2) T.Kind.k_effrow
+    (* contravariant *)
+    check_subtype env (T.Type.subst sub2 itp2) (T.Type.subst sub1 otp1);
```
forell: Typo: should use `itp1` instead of `otp1`.

forell (approved):
Looks good to me. I'll add an issue about implementing dynamic handlers in some form, since this PR removes labeled handlers.

These changes made me wonder about another possible modification of the handler type. Since those types are a bit like higher-order functions that take a function argument expecting a capability (kinda like in the translation to Core), maybe the capability type of a handler could be generalized to a scheme, just like how we can have schemes on the left-hand side of an arrow. Seems to me like it might work ok in case the capability pattern is actually just an identifier. What do you think?

PR #144, opened 2024-08-31
Adding flag to disable printing of error context

### LINKED ISSUE #142: Command line flag to disable "Better position printing"
In #113 we improved position printing in error messages. Now, error message contains piece of code providing context around the actual error. However, this feature breaks `@stderr:` directives supported by `test.sh` script, because the code around the error usually contains such a directive, and therefore, the condition set by the directive is trivially satisfied. The simplest solution to this problem is to add a command-line flag that disable this context printing, and use this flag in negative tests.

Foxinio:
Fixes #142

Also I added try...with segment in TextRangePrinting, because at this point we want to print error and it's location and if printing context fails we would rather see error's position without context, then information that printing context failed.

### DIFF
--- a/src/dbl.ml
+++ b/src/dbl.ml
@@ -47,7 +47,7 @@ let cmd_args_options = Arg.align
     Arg.String (fun p -> cli_local_search_dirs := p :: !cli_local_search_dirs),
     " Add a path to local search directories";
 
-    "--no-error-context",
+    "-no-error-context",
     Arg.Clear DblConfig.display_error_context,
     " Do not print piece of code with error, just filename:line:character";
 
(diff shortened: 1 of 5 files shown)


ppolesiuk (approved):
Looks great! Thanks!

src/dbl.ml
```diff
@@ -47,6 +47,10 @@ let cmd_args_options = Arg.align
     Arg.String (fun p -> cli_local_search_dirs := p :: !cli_local_search_dirs),
     " Add a path to local search directories";
 
+    "--no-error-context",
```
ppolesiuk, line 50: One minor thing: in `Arg` library, as well as in other OCaml tools, flags starts with single dash. I'm not happy with this practice, but since we use this library, we should respect it. I'll fix it before merge.

Issue #14, opened 2024-03-23
Some method definitions can lead to `Internal error: type mismatch`


ppolesiuk:
The following program exhibits the issue.
```
data U = I

method foo {self = I} = I

let baz I (x : U) = x.foo
```
Here’s the full error message.
```
Internal error: type mismatch
at:
(fn (x#16 tp#1b) (let x#1d (tapp foo#10 tp#8)) (app x#1d x#16))
requested:
(tp#1f -> tp#1f)
provided:
(tp#1f -> tp#1f (effect tp#8 tp#8))
Fatal error: exception Failure("Internal error: type mismatch")
```
Another example without pattern matching, forcing impurity through the annotation.
```
method foo {self : U} = (fn g => g : U ->[] U) (fn x => x) I
```

This issue was originally reported by @forell

PR #152, opened 2024-10-27
Pure pattern-matching

### LINKED ISSUE #149: Pure pattern-matching
It is well-known that recursive types are expressive enough to implement recursion and looping behavior. For instance, the following code falls into infinite loop, without using explicit recursion, except for types.
```
data rec Loop T = Loop of (Loop T ->[] T)
let loop (Loop loop) = loop (Loop loop)
let _ = loop (Loop loop)
```
In order to make sure that *pure* computations (e.g., those that are represented by `->` arrow) always terminate, we consider pattern-matching as an impure operation. When we change `Loop T ->[] T` into `Loop T -> T` in the example above, we would get a type error in the third line, because the `loop` function defined in the second line is impure, so it cannot be passed as the parameter of the `Loop` constructor.

The requirement that pattern-matching is impure sometimes is annoying, especially, when we work with non-recursive records. We could allow pure pattern-matching, when matched type is non-recursive or (strictly-)positively recursive. However, because of used representation of ADTs in Core, the change is not as simple as it seems.

To implement ADTs in Core we use special type `TData(tp, ctors)` that represents proofs that type `tp` is an ADT with constructors `ctors`, where `tp` can be an arbitrary type, so there is no direct way to check if the type is recursive or not. We could add additional parameter to `TData` constructor that signals when the ADT is non-positively recursive, and therefore the pattern-matching should be impure (this parameter can be represented as an effect of pattern-matching). Additionally, such an information could be stored in places where elements of `TData _` are created, i.e., in `DD_Data` constructor.

Similar changes could be made in `TypeInference.Module.adt_info` and `Lang.Unif.data_def`.

ppolesiuk:
This pull-request aims to resolve #149. This will be done in the following steps

- [x] Add flag about strictly positive recursion to definitions of ADT in Unif
- [x] Add flag about strictly positive recursion in ADT environment (`TypeInference.Module.adt_info`)
- [x] Recognition of strictly positive recursive types in `TypeInference`
- [x] Use information about strictly positive recursion in the effect of pattern-matching (`TypeInference.Pattern`)
- [x] Update translation to Core
- [x] Add flag about strictly positive recursion to definitions of ADT in Core
- [x] Add effect of pattern-matching to `TData` type in Core
- [x] Adjust internal type-checker for Core

### DIFF
--- a/src/Lang/Core.mli
+++ b/src/Lang/Core.mli
@@ -135,8 +135,13 @@ type data_def =
       args  : TVar.ex list;
         (** List of type parameters of this ADT. *)
 
-      ctors : ctor_type list
+      ctors : ctor_type list;
         (** List of constructors. *)
+
+      strictly_positive : bool
+        (** A flag indicating if the type is strictly positively recursive (in
+          particular, not recursive at all) and therefore can be deconstructed
+          without performing NTerm effect. *)
     }
 
   | DD_Label of (** Label *)
--- a/src/Lang/Core.mli
+++ b/src/Lang/Core.mli
@@ -84,14 +84,19 @@ type _ typ =
         (** Effect of the delimiter *)
     } -> ktype typ
 
-  | TData    : ttype * ctor_type list -> ktype typ
+  | TData    : ttype * effect * ctor_type list -> ktype typ
     (** Proof of the shape of ADT.
       
       Algebraic data type (ADTs) are just abstract types, but each operation
       on them like constructors or pattern-matching requires additional
       computationally irrelevant parameter of type that describes the shape
       of ADTs. This approach simplifies many things, e.g., mutually recursive
-      types are not recursive at all! *)
+      types are not recursive at all!
+
+      The element of type [TData(tp, eff, ctors)] is a witness that type [tp]
+      has constructors [ctors]. The effect [eff] is an effect of
+      pattern-matching: its pure for strictly positively recursive types,
+      and impure for other types (because it may lead to non-termination). *)
 
   | TApp     : ('k1 -> 'k2) typ * 'k1 typ -> 'k2 typ
     (** Type application *)
--- a/src/Lang/CorePriv/SExprPrinter.ml
+++ b/src/Lang/CorePriv/SExprPrinter.ml
@@ -54,8 +54,9 @@ let rec tr_type : type k. k typ -> SExpr.t =
         List (List.map tr_type lbl.val_types);
         tr_type lbl.delim_tp;
         tr_type lbl.delim_eff ]
-  | TData(tp, ctors) ->
-    List (Sym "data" :: tr_type tp :: List.map tr_ctor_type ctors)
+  | TData(tp, eff, ctors) ->
+    List (Sym "data" :: tr_type tp :: List (tr_effect eff) ::
+      List.map tr_ctor_type ctors)
   | TApp _ -> tr_type_app tp []
 
 and tr_effect : effect -> SExpr.t list =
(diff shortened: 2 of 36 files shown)


ppolesiuk:
This implementation has some slight problem with generated accessors for recursive records. For instance, when we write the following code
```
data rec  T = { x : T }
let f (t : T) = t.x
```
the `f` function is impure. This problem comes from the desugaring of recursive records. The above definition is a syntactic sugar for
```
rec
  data T = T of { x : T }
  method x { self = T { x } : T } = x
end
```
Notice, that the method is in recursive block, and therefore is impure, even if the pattern-matching on `self` is pure.

Foxinio:
All the changes look good to me, however I did look at the problem mention above and there seems to exist simple (partial) fix. It will make record accessors non recursive when record is marked `rec` and when it is in `rec` block with only other records, data definitions and labels. It won't always make accessors pure, like in this piece of code:
```
rec
  let foo (x : T) (n : Int) =
    if n == 0 then x
    else foo (x.x n) (n - 1)

  data T = { x : Int -> T }
end
```
If we ever decide to make non-termination detection more precise, for example like the one that's present in coq, my change will become redundant as whole problem will be solved. Until then it will help with the not obscure code.

Foxinio:
changes mentioned:
```
diff --git a/src/DblParser/Desugar.ml b/src/DblParser/Desugar.ml
index 27d4fe2..ed8bcae 100644
--- a/src/DblParser/Desugar.ml
+++ b/src/DblParser/Desugar.ml
@@ -42,6 +42,19 @@ let tr_ctor_name' (cname : Raw.ctor_name) =
 
 let with_nowhere data = { pos = Position.nowhere; data = data}
 
+let rec node_is_rec_data (def : Raw.def) =
+  match def.data with
+  | DRecord _ -> true
+  | DData   _ -> true
+  | DLabel  _ -> true
+  | _ -> false
+
+let node_is_data_def (def : def) =
+  match def.data with
+  | DData  _ -> true
+  | DLabel _ -> true
+  | _ -> false
+
 let annot_tp e tp =
   { pos = e.pos;
     data = Raw.EAnnot(e, with_nowhere tp)
@@ -751,6 +764,11 @@ and tr_def ?(public=false) (def : Raw.def) =
   | DOpen(pub, path) -> 
     let public = public || pub in
     [ make (DOpen(public, path)) ]
+  | DRec(pub, defs) when List.for_all node_is_rec_data defs ->
+    let public = public || pub in
+    let dds, accessors = tr_defs ~public defs
+      |> List.partition node_is_data_def in
+    make (DRec dds) :: accessors
   | DRec(pub, defs) ->
     let public = public || pub in
     [ make (DRec (tr_defs ~public defs)) ]
```

Foxinio (approved):

forell (approved):
I haven't spotted anything suspicious, so I'm in favor of merging. However, as (I think) @ilikeheaps pointed out, we should make sure existing tests still make sense after the change. I remember you identified [`ok0079_impureMethod.fram`](https://github.com/fram-lang/dbl/blob/0bc74238288f2a94b1be4f4355af976091151fd0/test/ok/ok0079_impureMethod.fram) as outdated, which tests the fix for an old bug reported in #14. We should force impurity in that test by some other means.

ppolesiuk:
@Foxinio These changes looks reasonable, but I think we should merge them in a separate pull-request. This PR propose basic solution to the problem of pure pattern-matching that has some limitations, but has no edge cases. Your fix improves the solution, but is not perfect, and in a long run will be probably replaced by something better. Having separate commits for them would make history cleaner, and it would be easier to revert your changes in order to provide the final solution.

PR #154, opened 2024-11-15
Stdlib List update

fatmage:
Added optional parameters to *i functions (mapi, iteri etc.) and tests for the whole List module.

### DIFF
--- a/lib/List.fram
+++ b/lib/List.fram
@@ -2,7 +2,7 @@
  * See LICENSE for details.
  *)
 
-pub let isEmpty xs =
+pub let isEmpty (xs : List _) =
   match xs with
   | [] => True
   | _  => False
@@ -23,7 +23,9 @@ pub let hd xs =
   end
 
 pub let hdErr { ~onError } xs =
+pub let hdErr { ~onError } xs =
   match xs with
+  | []      => ~onError ()
   | []      => ~onError ()
   | x :: xs => x
   end
@@ -35,7 +37,9 @@ pub let tl xs =
   end
 
 pub let tlErr { ~onError } xs =
+pub let tlErr { ~onError } xs =
   match xs with
+  | []      => ~onError ()
   | []      => ~onError ()
   | x :: xs => xs
   end
@@ -49,11 +53,14 @@ pub let nth xs (n : Int) =
   in if n < 0 then None else nthAux xs n
 
 pub let nthErr { ~onError } xs (n : Int) =
+pub let nthErr { ~onError } xs (n : Int) =
   let rec nthErrAux xs (n : Int) =
     match xs with
     | []      => ~onError ()
+    | []      => ~onError ()
     | x :: xs => if n == 0 then x else nthErrAux xs (n - 1)
     end
+  in if n < 0 then ~onError () else nthErrAux xs n
   in if n < 0 then ~onError () else nthErrAux xs n
 
 pub let rec append xs ys =
@@ -82,19 +89,23 @@ pub let rec map f xs =
   | x :: xs => f x :: map f xs
   end
 
-pub let mapi f xs =
-  let rec mapiAux (i : Int) f xs =
+pub let rec mapi {?i : Int} f xs =
+  match i with
+  | None   => mapi {i = 0} f xs
+  | Some i =>
     match xs with
     | []      => []
-    | x :: xs => f i x :: mapiAux (i + 1) f xs
+    | x :: xs => f i x :: mapi {i = i + 1} f xs
     end
-  in mapiAux 0 f xs
+  end
 
+pub let map2 { ~onError } f xs ys =
 pub let map2 { ~onError } f xs ys =
   let rec map2Aux xs ys =
     match xs, ys with
     | [],      []      => []
     | x :: xs, y :: ys => f x y :: map2Aux xs ys
+    | _,       _       => ~onError ()
     | _,       _       => ~onError ()
     end
   in map2Aux xs ys
@@ -108,10 +119,12 @@ pub let rec revMapAppend f xs ys =
 pub let revMap f xs = revMapAppend f xs []
 
 pub let revMap2 { ~onError } f xs ys =
+pub let revMap2 { ~onError } f xs ys =
   let rec revMap2Aux acc xs ys =
     match xs, ys with
     | [],      []      => acc
     | x :: xs, y :: ys => revMap2Aux (f x y :: acc) xs ys
+    | _,       _       => ~onError ()
     | _,       _       => ~onError ()
     end
   in revMap2Aux [] xs ys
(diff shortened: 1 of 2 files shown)


lib/List.fram
```diff
@@ -56,40 +68,55 @@ pub let nthErr { ~onError } xs (n : Int) =
     end
   in if n < 0 then ~onError () else nthErrAux xs n
 
+## `apppend xs ys` appends `ys` to `xs`.
```
wojpok, line 71: Maybe use word "concatanate"

lib/List.fram
```diff
@@ -56,40 +68,55 @@ pub let nthErr { ~onError } xs (n : Int) =
...
   end
 
+## `revAppend xs ys` reverses `xs` and appends it to `ys`
 pub let rec revAppend xs acc =
   match xs with
   | []      => acc
   | x :: xs => revAppend xs (x :: acc)
   end
 
+## Return the given list reversed.
```
wojpok, line 85: "Returns" sound better

lib/List.fram
```diff
@@ -192,24 +260,36 @@ pub let iter2 { ~onError } f xs ys =
...
+  match i with
+  | None    => init {i = 0} len f
+  | Some i =>
+    let rec initAux (i : Int) f =
+      if i >= len then []
+      else f i :: initAux (i + 1) f
+    in initAux i f
+  end
 
+## `foldLeft f init [x1, x2, ... ,xn]` is `f (... (f (f init x1) x2) ...) xn`.
```
wojpok, line 276: This explanation is not really useful

lib/List.fram
```diff
@@ -372,18 +489,20 @@ pub let mem
     end
   in memAux xs
 
+{## Check if two lists have the same lengths and their elements at equal
+  positions are equal. ##}
 pub let equal
   { E, method equal : E -> E ->[] Bool }
```
wojpok, line 495: Methods are really limited, for example this wont work for a `List List Int` type. I feel like we should include a variant that takes `eq` explicitely, so that comparing more elaborate types could be expressed. In the case of `List list Int` it could be written as `List.eq (List.eq Int.eq)` or something of that sort. Same logic can be applied to `toString` method.

lib/List.fram
```diff
@@ -400,17 +519,17 @@ pub method revAppend ys = revAppend self ys
...
+pub method filteri { ?i : Int } p = filteri { ?i = i } p self
 pub method filterMap f = filterMap f self
 pub method concatMap f = concatMap f self
 pub method drop n = drop n self
 pub method takeWhile p = takeWhile p self
 pub method dropWhile p = dropWhile p self
-pub method iteri f = iteri f self
+pub method iteri { ?i : Int } f = iteri { ?i = i } f self
 pub method foldLeft f acc = foldLeft f acc self
 pub method foldRight f acc = foldRight f self acc
```
wojpok, line 534: I would love to see a `foldRight1` and `foldLeft1` methods that takes last/first element of a list as an accumulator.

wojpok (commented):
In general I like those changes a lot. Most comments are fine, but would sound better if written in Present Simple tense (use "Returns" rather then "Return").

I feel like this might be a great opportunity to extend interface with some additional methods:
- foldRight1/foldLeft1 - folds that takes last/first element as an accumulator
- `equal`/`toString` functions variants that takes `equal`/`toString` function of a sub type explicitly.

lib/List.fram
fatmage: Concatenate already refers to and describes the ``concat`` function.

lib/List.fram
fatmage: In general I think we should discuss the standard library during one of our meetings to think what functionalities we are missing and would like to add. We could also unify the interfaces between similar modules e.g. ``List`` and ``Array``.

lib/List.fram
```diff
@@ -298,6 +480,12 @@ pub let rec findMap f xs =
     end
   end
 
+{##
+  Return sthe first value returned as `Some v` after application of `f` to an
+  element of the given list or calls `~onError` if there is no such result.
```
wojpok, line 485: Like above

lib/List.fram
```diff
@@ -310,30 +498,54 @@ pub let findMapErr { ~onError } f xs =
     end
   in findMapErrAux f xs
 
-pub let findMapi f xs =
-  let rec findMapiAux (i : Int) xs =
+{##
+  Same as `List.findMap` but it takes an additional optional parameter `i`
+  which is incremented with every consecutive application of `f`, which itself
+  takes and additional argument.
```
wojpok, line 504: Same issue as `mapi`

lib/List.fram
```diff
@@ -310,30 +498,54 @@ pub let findMapErr { ~onError } f xs =
...
+      | None    => findMapi {i = i + 1} f xs
       end
     end
-  in findMapiAux 0 xs
+  end
 
-pub let findMapiErr { ~onError } f xs =
-  let rec findMapiErrAux (i : Int) xs =
+{##
+  Same as `List.findMapErr` but it takes an additional optional parameter `i`
```
wojpok, line 523: Same issue as `mapi`

lib/List.fram
```diff
@@ -372,63 +590,68 @@ pub let mem
     end
   in memAux xs
 
+{##
+  Checks if two lists have the same lengths and their elements at equal
```
wojpok, line 594: Change "equal positions" to "coresponding positions"

wojpok (changes requested):
This PR is nearly done. I have noticed easy to fix issues with wording of some comments:

First and foremost, when dealing with function variants that taken indexing argument `i`, don't mention the possiblity of overriding initial `i` value in very first statement, for it is not a main feature of said functions. Just split the sentence and change order.

As to styling, I would suggest to replace `whose` with `which` for instance. It is not a mistake, but in formal setting like documentation, we really shouldn't personify inanimate objects.

I also strongly suggest to substitute `equal positons` with `corresponding positions`. It is more accurate.

test/stdlib/stdlib0003_List.fram
```diff
@@ -0,0 +1,284 @@
...
+
+let _ =
+  assert { msg = "foldLeftConcatMap" }
+    (let (v,xs) =
+      List.foldLeftConcatMap
+      (fn (acc : Int) (x : Int) => acc+x, List.init x (fn _ => x)) 0 xs in
+    v == 10 && xs == [1,2,2,3,3,3,4,4,4,4])
+
+let _ = assert { msg = "mem" } (List.mem {E=String} "b" ["a","b","c"])
+let _ = assert { msg = "mem" } (not (List.mem {E=String} "d" ["a","b","c"]))
```
ppolesiuk: No new line at the end of file. Unix dosn't like it.

PR #155, opened 2024-11-17
Int and Int64 conversion

fatmage:
Functions for converting Int to Int64 and vice versa were using opposite external functions.


ppolesiuk (approved):
Indeed. Now it looks better. Thanks!

Issue #130, opened 2024-06-16
Default `onError` instance in REPL


ppolesiuk:
It would be nice to have the default instance of `~re`/`~onError` in REPL. However the following naive definition:
```
data Error (effect E) = { raise : {type T} -> Unit ->[E] T }
handle {effect=RE} err = Error { raise = effect () / _ => () }
let ~onError = err.raise
```
have a huge drawback: when we type `[].hdErr`, the REPL would exit immediately. The problem can be solved at library level, by adding the following file (ReplUtils.fram) to the standard library:
```
pub data ReplUtils (effect E) =
  { raise  : {type T} -> Unit ->[E] T
  , commit : Unit ->[E] Unit
  }

pub handle {effect=ReplRE} repl = ReplUtils
  { raise  = effect () / _ => fn snapshot =>
    printStrLn "Runtime Error!";
    match snapshot with
    | None   => ()
    | Some f => f ()
    end
  , commit = effect () / r => fn _ =>
    let rec snapshot () = r () (Some snapshot) in
    snapshot ()
  }
  return  x => fn _ => x
  finally c => c None

let _ = repl.commit ()

pub let ~onError = repl.raise
```
but this would not solve all the problems:
```
> import List ;;
> import open ReplUtils ;;
> let foo x = x ;;
> [].hdErr ;;
: a
Runtime Error!
> foo 42 ;;
<stdin>:1:1-3: fatal error: Unbound variable foo
>
```
To do it correctly, we have to call `repl.commit` method after each REPL command, but this requires changes in the type-checker (or desugaring)

PR #158, opened 2024-11-22
Assertions and irrecoverable runtime errors

### LINKED ISSUE #157: Irrecoverable errors
We should add builtin function `runtimeError : {type T} -> String ->[] T` (or maybe someone could propose a better name?) for rising fatal, irrecoverable errors. Such a function can be used to implement `assert`, and used across standard library to fail in case of a major violation of function preconditions (e.g., division by zero).

ppolesiuk:
This commit introduces three functions to the standard library:

- `runtimeError` for raising fatal errors in impossible cases or in case of gross violation of function preconditions.
- `assertImpossible` for indicating impossible branches
- regular `assert` function

All these function are pure, so they should not be called at runtime of a correct program.

Resolves #157


src/Eval/External.ml
```diff
@@ -55,8 +55,12 @@ let int64_cmpop op = int64_fun2 (fun x y -> of_bool (op x y))
 
 let str_cmpop op = str_fun (fun s1 -> str_fun (fun s2 -> of_bool (op s1 s2)))
 
+let runtime_error msg =
+  failwith ("Runtime error: " ^ msg)
```
Foxinio, line 59: I don't think this should be implemented using `failwith`. Triggering this function will result in output:
```
Fatal error: exception Failure("Runtime error: Assertion failed")
```
Personally i think it should be Just `Runtime error: Assertion failed`, so this code should be changed to eprintf + exit 1

lib/Base/Assert.fram
```diff
@@ -0,0 +1,28 @@
...
+
+  Note that this function is pure. In correct programs, this function should
+  be never called at runtime, but it can be used to appease the type-checker.
+  Use this function only in impossible match clauses or in case of gross
+  violation of function preconditions (e.g., division by zero). *)
+pub let runtimeError {type T} =
+  (extern dbl_runtimeError : String ->[] T)
+
+(** Explicitly assert, that this case is impossible *)
+pub let assertImpossible {?msg : String} () =
```
Foxinio, line 17: I would probably change this to just `impossible`, even more since it's in module Assert.

Foxinio (commented):
Marked two minor changes i would prefer, but aside from that all seems good. Maybe if we decide to add reflection (like macros evaluating to filename, line number, etc) this module could be revisited.

src/Eval/External.ml
MinionJakub: I would add possibility to add custom error code/number on exit.

ppolesiuk: I agree that this error should not be implemented with `failwith`. Maybe a custom exception that is handled in the main file would be better? The REPL implementation could catch this exception and recover with a prompt for an another command, instead of aborting the interpreter.

Foxinio: Actually that would probably be best

MinionJakub: I don't know and possible I am wrong but this sounds like an error implementation which we would make in standard library.

ppolesiuk: Do you mean the issue #130? I think these two are two different things. In this PR we talk about errors that should never occur in correct programs (but unfortunately, most of programs are incorrect), so the effect of such error is pure (`[]`). In #130 the error has some effect visible in types. However, after merging this PR, we may rethink if we still want to implement #130.

ppolesiuk: However, I realized that catching runtime error in REPL is not as trivial as I thought. It is not just adding extra match clasue in [Eval.ml](https://github.com/fram-lang/dbl/blob/f54c429d0f20614d0d91a616908f7cbd329ce1e6/src/Eval/Eval.ml#L118), because the exception are catch before the evaluation. I think it is still possible to handle it correctly, because the evaluator is in CPS, but it would be a more complex change, and should be done in a separate issue.

Foxinio (commented):
One thing to fix, and in my opinion PR is ready to merge

forell:
I brought this PR up to date with master and made some small changes based on the discussion we had.
* I got rid of the program-ending version of `Int64.toInt` and made the `Option` variant the default. The failing version could be reintroduced as `toIntUnsafe` or something, but I think it's nicer to minimize the number of such "unsafe" functions, so...
* I added `Base.Option` to the standard library, with two methods to unwrap optionals: `unwrap` (unsafe, uses `runtimeError`) and `unwrapOr` (takes a default value as an argument). I'm open to other names for these methods, I just went with what Rust uses. Also, `unwrapOr` is similar to the various implicit-based `*Err` functions that exist in the standard library, except the argument is not named and not a function. The proposed version *is* convenient for short expressions like `n.toInt.unwrapOr 0`, though.  
* The `get` method for strings remains unsafe. Littering code using indexing with `unwrap`s or long method names would be too noisy when it's normally used in a context where the bounds have already been checked by the programmer. Safe variants that allow for recovery could be introduced.

lib/Base/Option.fram
ppolesiuk, line 1: I agree with these two methods, and proposed names are fine. I think, the `unwrapErr` function is missing. As a use case, consider that we have some polymorphic `~onError` implicit in the context, so we can write a code optimistically using `*Err` functions without explicit error handling, but we still can recover from an error in some handler.

Moreover, other methods and functions for manipulating `Option` type, like `map`, `fold`, `iter`, etc. are missing, but the can be added as a separate issue.

lib/Base/Option.fram
```diff
@@ -0,0 +1,25 @@
+{# This file is part of DBL, released under MIT license.
+ # See LICENSE for details.
+ #}
+
+import open /Base/Types
+import open /Base/Assert
+
+{## For `Some x` returns `x`, otherwise the provided argument is returned.
+ #}
+pub method unwrapOr default =
```
ppolesiuk, line 10: I was wondering, if the `default` parameter should be named, because in almost all cases it will be provided. But I think, the proposed interface is good: the name of this method suggests what is the meaning of its first parameter, and writing code like `x.unwrapOr {default = 42}` would make some unnecessary noise.

ppolesiuk (commented):
I agree with the changes, and generally, I approve this PR (this PR is created be me, so GitHub forbids me from approving it formally). I have some minor comments (below), that I would like to have discussed, but not necessarily implemented.

MinionJakub (approved):
It looks good.

lib/Base/Option.fram
Foxinio: Alternative (although I don't know how good) would be for `unwrap` to take optional argument default.

lib/Base/Int.fram
```diff
@@ -19,13 +20,13 @@ pub method add = (extern dbl_addInt : Int -> Int -> Int) self
 pub method sub = (extern dbl_subInt : Int -> Int -> Int) self
 pub method mul = (extern dbl_mulInt : Int -> Int -> Int) self
 
-pub method div {~re : {type X} -> Unit ->[|_] X} (n : Int) =
-  if n.equal 0 then ~re ()
-  else (extern dbl_divInt : Int -> Int -> Int) self n
+pub method div (n : Int) =
+  assert {msg="Division by zero"} (n != 0);
+  (extern dbl_divInt : Int -> Int -> Int) self n
```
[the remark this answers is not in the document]
Foxinio: Well the effect in this case doesn't come from the extern, but from assertion, so in my opinion it works both ways.

Foxinio (approved):
I agree that more `Option` functionality should be added, and can be done in separate PR. And beside that, all changes look good and should be merged.

lib/Base/Option.fram
ppolesiuk: Functions `unwrap` with optional parameter and `unwrapOr` have different effects. The latter is more pure (`->` instead of `->[]`).

PR #159, opened 2024-11-23
Documentation of the internal Core type checker.

### LINKED ISSUE #151: Better documentation of `Lang.CorePriv.WellTypedInvariant`
Keeping this file relatively simple and well-documented is very important, because currently it is the only specification of Core type system. However, some nontrivial functions in this module are not documented.

ppolesiuk:
Fixes #151


forell (approved):
One comment seems to have the behavior of its arguments swapped compared to the implementation, so double check that and fix it if necessary before merging. Otherwise, everything in this file appears to be well-documented now.

PR #160, opened 2024-11-24
Changes to rec records

Foxinio:
Changes mentioned in #149, they should make it so that most of the time accessors to record fields marked `data rec...` aren't unnecessarily marked as impure.
1. It's not gonna always work, by which i mean that sometimes they will be marked as impure even though they aren't, like in this code presented in original PR:
```
rec
  let foo (x : T) (n : Int) =
    if n == 0 then x
    else foo (x.x n) (n - 1)

  data T = { x : Int -> T }
end
```
2. It's gonna become obsolete if we decide to make non termination more precise.

### DIFF
--- a/src/DblParser/Desugar.ml
+++ b/src/DblParser/Desugar.ml
@@ -42,6 +42,19 @@ let tr_ctor_name' (cname : Raw.ctor_name) =
 
 let with_nowhere data = { pos = Position.nowhere; data = data}
 
+let rec node_is_rec_data (def : Raw.def) =
+  match def.data with
+  | DRecord _ -> true
+  | DData   _ -> true
+  | DLabel  _ -> true
+  | _ -> false
+
+let node_is_data_def (def : def) =
+  match def.data with
+  | DData  _ -> true
+  | DLabel _ -> true
+  | _ -> false
+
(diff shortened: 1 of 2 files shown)


src/DblParser/Desugar.ml
```diff
@@ -751,6 +764,11 @@ and tr_def ?(public=false) (def : Raw.def) =
   | DOpen(pub, path) -> 
     let public = public || pub in
     [ make (DOpen(public, path)) ]
+  | DRec(pub, defs) when List.for_all node_is_rec_data defs ->
```
ppolesiuk, line 767: I mean, a comment describing this edge case.

ppolesiuk (commented):
I think, these changes can be seen as a quick fix, and in a long run we will change them to a more systematic, but more complex solution. Therefore, I would like to see a short comment describing what's going on.

src/DblParser/Desugar.ml
Foxinio: Agreed

ppolesiuk:
Do we want to have some tests?

Foxinio:
It won't hurt to have test for that

Issue #162, opened 2024-12-03
New syntax of comments


ppolesiuk:
We agreed on the following syntax of comments in Fram (current syntax is inconvenient, in case of defining `*` operator):
```
# this is a single-line comment
{#
  multi-line
  comment
#}
```
The `#` character should be also removed from the list of allowed operator characters in lexer.

PR #164, opened 2024-12-08
changed comment syntax as requested 

dominik-muc:
from #162

I hope I didn't miss anything...


ppolesiuk (approved):
Thanks for your contribution! The changes looks good. Two minor things should be fixed:

1. While the character `#` is removed from the function deciding priority of the operator (`tokenize_oper`), it is still recognized as an operator character by the lexer (see `op_char` regex).
2. Regular expressions for recognizing `@stderr` and `@stdout` directives in the testing script (`test.sh`) should be fixed.

I'll not bother you, and I'll fix these two trivial things myself. Thanks!

PR #165, opened 2024-12-10
Replace ocamlyacc with cpspg

adampsz:
This PR replaces ocamlyacc with adampsz/cpspg@67dfc0f3190a06363292c14c9315dd9c7dbc120b to check its compatibility and experiment with the new parser generator tool.

Building now requires `cpspg` to be in the path - it can be `dune build` and `dune install`-ed from adampsz/cpspg repo.

### DIFF
--- a/src/DblParser/dune
+++ b/src/DblParser/dune
@@ -1,5 +1,10 @@
-(ocamlyacc (modules YaccParser))
 (ocamllex  (modules Lexer))
+;(ocamlyacc (modules YaccParser));
+
+(rule
+ (deps YaccParser.mly)
+ (target YaccParser.ml)
+ (action (run cpspg --compat -o %{target} %{deps})))
 
 (library
   (name dblParser)
--- a/src/DblParser/YaccParser.mly
+++ b/src/DblParser/YaccParser.mly
@@ -145,7 +145,7 @@ op
 | op_80  { $1 }
 | op_90  { $1 }
 | op_100 { $1 }
-; 
+;
 
 /* ========================================================================= */
 
@@ -391,7 +391,7 @@ expr_90
 ;
 
 // exp1 ** exp2
-expr_100 
+expr_100
 : expr_150 op_100 expr_100 { make (EBOp($1, $2, $3)) }
 | expr_150 { $1 }
 ;
@@ -417,6 +417,7 @@ expr_select
 : UID DOT expr_ctor   { (NPName $1, $3) }
 | UID DOT expr_300    { (NPName $1, $3) }
 | UID DOT expr_select { let (p, e) = $3 in (NPSel($1, p), e) }
+;
 
 expr_250
 : expr_300    { $1 }


src/DblParser/dune
```diff
@@ -1,5 +1,10 @@
-(ocamlyacc (modules YaccParser))
 (ocamllex  (modules Lexer))
+;(ocamlyacc (modules YaccParser));
```
ppolesiuk, line 2: Should we remove this line instead of commenting it out? As discussed on the last meeting, we are going to use nice features of cpspg (which are not implemented yet), so I think in the near future, we will be incompatible with ocamlyacc.

src/DblParser/dune
```diff
@@ -1,5 +1,10 @@
...
+
+(rule
+ (deps YaccParser.mly)
+ (target YaccParser.ml)
```
ppolesiuk, line 6: Does cpspg generate `.mli` files? If yes, maybe it should be added to this rule (I'm sure, dune can do that. You can ask @Foxinio, our local dune guru). If not, maybe we could consider writing own `YaccParser.mli` file.

ppolesiuk (changes requested):
Thanks! This PR opens new perspectives for a better parser. However, I cannot merge it yet, since it breaks CI.

Brychlikov:
I got CI to work using nix [here](/Brychlikov/dbl/tree/cpspg). I packaged both `dbl` and `cpspg` as flakes.
For now, I had to make two changes: 
- `test.sh` doesn't try to run `dune build`
- changed its first parameter so it now should be something in `PATH` or be a path to `dbl` binary

flake.nix
```diff
@@ -0,0 +1,45 @@
+{
+  inputs = {
+    opam-nix.url = "github:tweag/opam-nix";
+    flake-utils.url = "github:numtide/flake-utils";
+    nixpkgs.follows = "opam-nix/nixpkgs";
+    cpspg.url = "github:Brychlikov/cpspg";
```
adampsz, line 6: adampsz/cpspg#1 was merged, so this can be replaced with `cpspg.url = "github:adampsz/cpspg"`

ilikeheaps: Github has this cool feature for suggestions (works with multiline comments too):
```suggestion
    cpspg.url = "github:adampsz/cpspg";
```

test.sh
```diff
@@ -1,18 +1,18 @@
...
+# if ! dune build; then
+# 	exit 1
+# fi
 
 export DBL_LIB="lib/"
 
 TIMEOUT=1.0
 
-binary="_build/default/src/$1.exe"
+binary=$1
```
ppolesiuk, line 15: I see a minor problem with this change. When I run tests locally, I would like to use a version of dbl compiled from source on the active branch, not the one that is installed on my system.

src/DblParser/dune
adampsz: It does not generate `.mli` files, but it should be straightforward to implement.

test.sh
Brychlikov: A bit of an oversight on my part - in my dev environment, `dbl` in `PATH` is the one from `_build/...`. To be fair, you still can run `./test.sh ./_build/default/src/dbl.exe ./test/test_suite` to test against your freshly built `dbl`, but I understand that the default workflow should probably remain unchanged.

Brychlikov: I restored most of the original behaviour of `test.sh` in my branch - when passed a relative path as `PROGRAM` it does what it used to do.

PR #168, opened 2024-12-14
New implementation of block commets

### LINKED ISSUE #166: Awkward interaction of doc comments and single-line comments
The convention for documentation comments that we agreed on a while back is something like follows.
```
{## A comment. #}
```
But, because `#` also starts single-line comments, if the closing `#}` is one the same line as the extra `#`, it won't get picked up by the lexer. This will also come up if we try to use markdown headers in comments.

ppolesiuk:
The behavior of block comments is changed, as proposed in the documentation of Fram.

We can wait with merging this PR, until we agree on the approach to block comments. But I hope, this PR shows that the solution proposed in #166 has a simple implementation.

Fixes #166


forell (approved):
Looks good, and I appreciate how even the implementation is simpler than with the previous approach (which was missing various edge cases anyway). Are we still waiting for feedback before merging?

ppolesiuk:
I think, I was loud enough on discord, and we've got positive feedback on Friday. But I would still wait, until you accept (or reject?) another [pull-request](https://github.com/fram-lang/fram-doc/pull/6) to the documentation. If it will be accepted, I can change the implementation.

PR #172, opened 2024-12-21
Line directives for the lexer

### LINKED ISSUE #128: Line directives for the lexer
We should support line directives for the lexer in order to have better error messages in a generated code. We will use them in [cpspg](https://github.com/fram-lang/cpspg). Of course, we have to agree on syntax.

ppolesiuk:
This PR introduces line directives recognized by the lexer. For instance, in
```
#@ 123 foo.fram
let x = ... some error ...
```
the error will be reported in `foo.fram` at line 123.

Moreover, the lexer warns when it encounters other comments starting with `#@`. We might use such comments for other directives in the future.

Resolves #128

### DIFF
--- a/src/DblParser/Error.ml
+++ b/src/DblParser/Error.ml
@@ -50,6 +50,14 @@ let invalid_escape_code pos =
 let eof_in_string pos =
   (Some pos, "Unexpected end of file inside a string literal")
 
+let invalid_lexer_directive ?msg pos =
+  (Some pos,
+    Printf.sprintf
+      "Invalid lexer directive%s"
+        (match msg with
+        | None -> ""
+        | Some msg -> ": " ^ msg))
+
 let desugar_error pos =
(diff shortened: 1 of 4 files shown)


Brychlikov:
My only worry is the legibility of error messages. Currently, if there is an error with line directive pointing to a nonexistent file or a nonexistent line in an existing file, we just quietly don't show any context. I think there should at least be a warning in this case. 
I don't have much experience working with code generators, so other than that, I think it's fine.

adampsz (commented):
Wouldn't it be worth using string literals as file names? This way, there wouldn't be any unsupported file names, and non-standard characters (e.g., spaces) would not be confusing. OCaml (https://ocaml.org/manual/5.2/lex.html#sss:lex-linedir) and GCC (https://gcc.gnu.org/onlinedocs/cpp/Line-Control.html) use this approach.

PR #174, opened 2024-12-22
New Surface language

ppolesiuk:
This is my proposition of the new Surface language. The main changes includes:

- Changed representation of module paths: the order on a list is reversed. I think it would be more convenient during reporting good error messages).
- Removed `effect` type name. With the new effect inference it will be not needed.
- Removed `label` name. I think it should be changed to some implicit during desugaring phase.
- Removed `effrow` kind.
- Removed row from `TEffect` construct.
- Added abstract syntax for handler and label types.
- Changed the representation of constructor visibility. The flag is not attached to the constructor declaration, but to ADT definition (there are two flags there).
- Named patterns unifies type and value patters. There are two other constructors `NP_Module` and `NP_Open`.
- Explicit instantiation unifies type and value instantiations (similarly to named patterns).
- There are two constructors for an application: to an expression, and to a polymorphic function (e.g., `e1 (fn {x} => e2)`).
- `EEffect` construct takes optional label parameter.
- `is_public` flag is detached from `ident` type (I'm not sure, if it is the right choice).
- `DImplicit` is changed into more general `DTypeParam` and `DValParam`. There is no reminiscent of the second parameter of `DImplicit`, but it can be emulated via `DTypeParam` if necessary.
- Introduced `DBlock` definition.

I didn't touch the parser. It could be done in a separate PR.

### DIFF
--- a/src/Lang/Surface.ml
+++ b/src/Lang/Surface.ml
@@ -199,7 +199,7 @@ and named_pattern_data =
   | NP_Open
     (** Introduce everything into the environment *)
 
-(** Polymorphic expressions *)
+(** Polymorphic expressions, at the place of use. *)
 type poly_expr_use = poly_expr_use_data node
 and poly_expr_use_data =
   | EVar      of var path
@@ -211,7 +211,7 @@ and poly_expr_use_data =
   | EMethod   of expr * method_name
     (** Call of a method *)
 
-(** Polymorphic expression *)
+(** Polymorphic expressions, at the place of definition. *)
 and poly_expr_def = poly_expr_def_data node
 and poly_expr_def_data =
   | PE_Expr of expr
--- a/src/Lang/Surface.ml
+++ b/src/Lang/Surface.ml
@@ -149,7 +149,7 @@ and scheme_arg_data =
   | SA_Val  of name * scheme_expr
     (** Value parameter *)
 
-(** Type formal parameter *)
+(** Formal type parameter *)
 and type_arg = type_arg_data node
 and type_arg_data =
   | TA_Var of tvar * kind_expr
@@ -158,6 +158,7 @@ and type_arg_data =
   | TA_Wildcard
     (** Wildcard *)
 
+(** Named formal type parameter *)
 and named_type_arg = (tname * type_arg) node
 
 (** Declaration of constructor of ADT *)
--- a/src/Lang/Surface.ml
+++ b/src/Lang/Surface.ml
@@ -25,24 +25,23 @@ type method_name = string
 type module_name = string
 
 (** Module path to an identifier of type 'a *)
-type 'a path =
+type 'a path = 'a path_data node
+and 'a path_data =
   | NPName of 'a
-  | NPSel  of module_name * 'a path
+  | NPSel  of module_name path * 'a
 
 (** Extract the base name from a path *)
-let rec path_name = function
-  | NPName x    -> x
-  | NPSel(_, p) -> path_name p
+let path_name (path : _ path) =
+  match path.data with
+  | NPName x | NPSel(_, x) -> x
 
 (** Name of a named type parameter *)
 type tname =
   | TNAnon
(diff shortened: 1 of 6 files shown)


src/Lang/Surface.ml
```diff
@@ -87,21 +82,55 @@ and type_expr_data =
...
   | TApp of type_expr * type_expr
     (** Type application *)
 
 (** Type-scheme expressions *)
 and scheme_expr = {
-  sch_pos : Position.t;
+  sch_pos   : Position.t;
     (** Location of the scheme expression *)
 
   sch_targs : named_type_arg list;
```
forell, line 136: I think forcing named types to come before named value parameters in schemes and constructor declarations will seem inconsistent now that this restriction is lifted in patterns.

For example, I believe the following uses of similar-looking syntax should have similar behavior regarding binding visibility, but I'm not sure they will if Surface is defined like this.
```fram
data T = ...

data U = C of {x : T, T} # Same as {T, x : T}?

# Referring to the previously defined T for x instead?
let C {x : T, T} = ...
```

src/Lang/Surface.ml
```diff
@@ -134,7 +160,6 @@ and named_type_arg = (tname * type_arg) node
 (** Declaration of constructor of ADT *)
 type ctor_decl = ctor_decl_data node
 and ctor_decl_data = {
-  cd_public      : is_public;
   cd_name        : ctor_name;
   cd_targs       : named_type_arg list;
   cd_named       : named_scheme list;
```
forell, line 165: As above.

src/Lang/Surface.ml
```diff
@@ -163,8 +188,17 @@ and ctor_pattern_named =
   | CNModule of is_public * module_name
```
forell, line 188: The type `ctor_pattern_named` is no longer used and can be removed.

src/Lang/Surface.ml
```diff
@@ -247,15 +282,24 @@ and expr_data =
...
+    (** Explicit instantiation, that takes values from the environment *)
 
 (** Local definitions *)
 and def = def_data node
 and def_data =
-  | DLetId of ident * expr
+  | DLetId of is_public * ident * expr
     (** Let definition: monomorphic or polymorphic, depending on effect *)
 
-  | DLetFun of ident * named_type_arg list * named_arg list * expr
```
forell, line 258: Polymorphic lambdas in argument position allow any order of type/named value parameters, so I think so should let-definitions for consistency.

src/Lang/Surface.ml
```diff
@@ -264,24 +308,39 @@ and def_data =
...
+
+      public_ctors : is_public;
+        (** A flag indication that the constructors are visible *)
+
+      args         : named_type_arg list;
+        (** List of type parameters *)
+
+      ctors        : ctor_decl list
+        (** List of constructors *)
+    }
```
forell, line 340: Where did the `tvar` go?

forell (commented):
Overall I agree with these changes, but there are a few things that went unchanged requiring some additional discussion, mainly to do with places where type parameters still have to come before named value parameters.

src/Lang/Surface.ml
ppolesiuk: Polymorphic lambdas are written in a context, where the scheme is known. Allowing any `named_pattern` here, leads to some problematic cases, like:
```
let foo {module M} = M.bar M.baz
```
But maybe such programs can be rejected by the type-checker.

src/Lang/Surface.ml
```diff
@@ -223,6 +211,18 @@ and poly_expr_data =
   | EMethod   of expr * method_name
     (** Call of a method *)
 
+(** Polymorphic expression *)
```
forell, line 214:
```suggestion
(** Definition of a polymorphic expression *)
```

forell (approved):
Looks good to me now.

PR #175, opened 2024-12-23
Added missing constructor

ppolesiuk:


forell (approved):
For context, this constructor was removed back in #41, presumably because the environment needed to disambiguate operators as values, methods or constructors anyway.

ppolesiuk:
That's true. I forgot about it. So maybe we should not merge this PR.

PR #179, opened 2025-01-11
Changes in Unif and Surface required by the new type inference

ppolesiuk:
These changes were required by the new type inference. I send this PR before I finished the type inference, because it would allow to start work on the next phase of the pipeline.


forell (approved):
Looks reasonable.

Issue #143, opened 2024-08-28
Checking types of constructors defined in Prelude


ppolesiuk:
Some constructors are defined in Prelude, but used by some transformations (see `TypeInference.PreludeTypes`). Currently, the implementation doesn't check their types: it assumes that they are defined as usual. In extremely rare cases, these assumptions are not correct, which leads to "Internal type error" message, like in the following code:
```
data Option X = None | Some

let foo {?x} = x

let _ = foo {x=42}
```
I'm not sure if this is a right design. Handling such cases correctly would require some complicated code, just to report an error that should be never seen by the casual programmer. The "Internal type error" in such rare cases seems to be good enough. On the other hand, I treat internal type errors as a tool for developers: when they occur, something is really bad in the DBL implementation. Assuming that our implementation is correct, the user should be not able to see such an error.

ppolesiuk:
Resolved by #194

Issue #119, opened 2024-06-04
Consistent rules of automatic binding of types and values in patterns


ppolesiuk:
If a constructor pattern or a polymorphic actual parameter does not mention some named parameter, it could be introduced to the environment or not. Currently implemented rules are not very consistent:
```
data T = C of {`n : Int}
let f C = `n (* is valid, because `n was implicitly introduced *)
let g (C { module M }) = `n (* Unbound implicit `n *)
```
nor intuitive:
```
let f (g : ({x} -> _) -> _) = g x (* is valid, since x was implicitly introduced *)
```
We should develop more clear, consistent, and intuitive rules how this feature should behave.

Issue #98, opened 2024-05-13
Refactor the implementation of environments and implicit environments


ppolesiuk:
Current implementation of implicit environment doesn't scale well into implicit declaration shadowing in presence of recursion. The following changes could be made.

- Merge (most of) functionality of `Env` and `ImplicitEnv`. It should simplify the management of implicit declaration and allow to get rid of name maps returned by `Pattern.check_type` (or at least these maps would not be necessary to implement shadowing; they still might be useful for implementing or-patterns).
- The environment could be responsible for detecting multiple definitions of the same name within e.g. block of recursive definitions. Current implementation based on syntactic checks (the `Uniqueness` module) is not able to detect multiple definitions of the same method. These check should be more semantic, because methods of the same name, but for different types should be considered different.

Issue #92, opened 2024-05-07
Opening ADT constructor as a module


ppolesiuk:
A module could be passed as a named parameter to ADT constructor pattern. Such a construct would define a new local module containing all named parameters. The motivating example is to use encode kind of generative functors known from OCaml or SML, using just named parameters.
```
data Map Key = Map
  { T
  , empty : {Val} ->[] T Val
  , method add : {Val} -> T Val -> Key -> Val ->[] T Val
  , method find : {Val} -> T Val -> Key ->[] Option Val
  }

(* Functor-like function *)
let makeMap {Key} (compare : Key -> Key ->[] Ord) =
  Map { ... }

(* Functor-like application: creating a new module StrMap that implements maps *)
let (Map { module StrMap }) = make String.compare
```

PR #180, opened 2025-01-13
New type inference

ppolesiuk:
The new type inference is almost ready. I've refactored lot of code, fixing some known issues and preparing the ground for the new effect inference. The list of fixed issues includes:
- issue #10 
- issue #11 (some decisions were made and I'm not sure that they are right)
- issue #32 
- issue #33
- issue #50 
- issue #51
- issue #59
- issue #92
- issue #119 (maybe?)

I also prepared some infrastructure for addressing the following issues
- issue #3. I pass `app_type` optional parameter that stores type type of the whole application/variable expression to some functions of the type-checker, especially `PolyExpr.infer_use_scheme`. Taking the head of whnf form of this type allows to lookup for ADT definition in the environment. However, this requires some changes in the Surface, because, we cannot distinguish variables from constructor now. The problem is not so easy, because one can define binary operators as constructors and its hard to distinguish them from regular operators. One possible solution that I see is to implement this checking-mode of constructors only for non-operator constructors.
- issue #20. Now the type-checker passes the constraints around. However the constraints and their solving are not implemented yet.
- issue #34. The task is delegated to some `Utils.PPTree` module (not implemented yet).
- I would like to get rid of "method functions" and issue #57. With delayed method constraints, method functions can be represented as regular functions.

Other important changes are summarized as follows.
- `Option` type is a built-in type, like `Unit`. Therefore, I think issue #143 doesn't make sense anymore.
- I get rid of type hints (the most ugly heuristics in the type-checker)
- Type-checking of pattern-matching returns partial environments. This approach paves the path for implementing or-patterns.
- Functionality of `ModStack` was moved to `Env` (one level of indirection less...).
- I've implemented general `parameter` mechanism.
- Resolving of named parameters was moved from `ExprUtils` to `ParamResolve`.
- I've created a separate module for checking of explicit instantiations (`Inst`)

In order to finish this type inference, the following thing should be implemented:
- [ ] Implement `PPTree` in order to make pretty-printer working
- [ ] Implement recursion
- [ ] Implement method constraints
- [ ] Fix a bug with annotation of optional parameters. This probably will require some changes in the Surface, so I would like to ask @forell for help.
- [ ] Implement visibility of `open` and `module` named patterns (this will also require changes in the Surface)
- [ ] Fix other bugs
- [ ] Refactor representation of names a bit. Wile names as defined in `Unif` works well in types, in many other places it is convenient to have a representation of names, where a method name contains the owner. This approach would allow to implement explicit method parameters, and make code cleaner in some places. Moreover, introducing a special type for method owner would allow to define methods for functions, handlers, and labels.
- [ ] Move most of functionality from `ParameterEnv` to `Env` (issue #98). I would like to get rid of the whole "shadowing circus" in `ParameterEnv`.
- [ ] Store `PPTree`s in Unif AST in order to get nice pretty-printing of error messages in the next phase.

Not all of them should be done in this PR.


ppolesiuk:
I've changed the Unif IR a bit. The most important changes include the following:
- Constructors are not fully-applied
- Recursive blocks separates parameters implicitly generalized in all definitions from other parameters. Productivity checking doesn't add local definitions, but marks the place where they should be added by `ERecCtx` construct.
- Function parameters are annotated with `scheme_expr` instead of `scheme`. Or they are not annotated at all, when the type is known from the context.
- AST nodes contain pretty-printing context (`PPTree.t`)
- Some definitions, that will be shared with ConE and EffectInference were moved to `Utils` or the new library `UnifCommon`.

I hope, the new Unif is better suited for the effect inference now, so I mark this PR as "ready for review".

PR #181, opened 2025-01-15
Optional param annotations & module pattern visibility

forell:
Two Surface-related changes requested in #180.
* To deal with optional parameter annotations, the `NP_Val` constructor of named patterns now contains a scheme annotation which is treated in a special way.
* The `open` and `module M` named patterns now store their visibility. I also made a small change in `PartialEnv` because the contents of modules were introduced with the wrong visibility in `introduce_module`, causing the relevant test to fail.

### DIFF
--- a/src/DblParser/Desugar.ml
+++ b/src/DblParser/Desugar.ml
@@ -343,37 +343,36 @@ let rec tr_pattern ~public (p : Raw.expr) =
   | EMethod _ | EExtern _ | EIf _ | EMethodCall _  ->
     Error.fatal (Error.desugar_error p.pos)
 
-(** Translate a pattern, separating out its annotation if present or creating
-  a wildcard annotation otherwise. *)
+(** Translate a pattern, separating out its annotation [Some sch] if present
+    or returning [None] otherwise. *)
 and tr_annot_pattern ~public (p : Raw.expr) =
   match p.data with
   | EParen p       -> tr_annot_pattern ~public p
-  | EAnnot(p, sch) -> tr_pattern ~public p, tr_scheme_expr sch
+  | EAnnot(p, sch) -> tr_pattern ~public p, Some (tr_scheme_expr sch)
   | EWildcard | EUnit | EVar _ | EBOpID _ | EUOpID _ | EImplicit _ | ECtor _
   | ENum _ | ENum64 _ | EStr _ | EChr _ | EFn _ | EApp _ | EDefs _ | EMatch _
   | EHandler _ | EEffect _ | ERecord _ | EMethod _ | EMethodCall _ | EExtern _
   | EIf _ | ESelect _ | EBOp _ | EUOp _ | EList _ | EPub _ ->
-    tr_pattern ~public p, scheme_wildcard p.pos
+    tr_pattern ~public p, None
 
 and tr_named_pattern ~public (fld : Raw.field) =
   let make data = { fld with data = data } in
   match fld.data with
-  | FldAnonType _ ->
-    Error.fatal (Error.anon_type_pattern fld.pos)
+  | FldAnonType arg ->
+    make (NP_Type(false, make (TNAnon, tr_type_arg arg)))
   | FldType(x, ka) ->
     let k = Option.value ka ~default:(make KWildcard) in
     make (NP_Type(public, make (TNVar x, make (TA_Var(x, k)))))
   | FldTypeVal(x, arg) ->
     make (NP_Type(public, make (TNVar x, tr_type_arg arg)))
   | FldName n ->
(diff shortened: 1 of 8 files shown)


src/DblParser/Desugar.ml
```diff
@@ -354,16 +366,17 @@ and tr_named_pattern ~public (fld : Raw.field) =
   | FldTypeVal(x, arg) ->
     make (NP_Type(public, make (TNVar x, tr_type_arg arg)))
   | FldName n ->
-    make (NP_Val(n, make (PId(public, ident_of_name n))))
+    make (NP_Val(n, make (PId(public, ident_of_name n)),
+                 scheme_wildcard fld.pos))
```
ppolesiuk, line 370: Assigning wildcard pattern here enforces that the parameter is monomorphic. However, we should allow polymorphic parameters without any annotation, when we are in the scheme-checking mode. For instance, consider the following program.
```
data T = T of {id : {X} -> X -> X}
let foo (T {id}) = id id 42
```
This type-checker produces the following error:
```
test.fram:2:13-14: error: Annotating pattern with type a, but it was expected to match values of type {X} -> X -> X
test:fram:2:23-24: error: This expression has type a ->> b, but an expression was expected of type a
```
I think, `NP_Val` constructor should have an optional annotation.

src/DblParser/Desugar.ml
```diff
@@ -343,6 +343,18 @@ let rec tr_pattern ~public (p : Raw.expr) =
...
+  a wildcard annotation otherwise. *)
+and tr_annot_pattern ~public (p : Raw.expr) =
+  match p.data with
+  | EParen p       -> tr_annot_pattern ~public p
+  | EAnnot(p, sch) -> tr_pattern ~public p, tr_scheme_expr sch
+  | EWildcard | EUnit | EVar _ | EBOpID _ | EUOpID _ | EImplicit _ | ECtor _
+  | ENum _ | ENum64 _ | EStr _ | EChr _ | EFn _ | EApp _ | EDefs _ | EMatch _
+  | EHandler _ | EEffect _ | ERecord _ | EMethod _ | EMethodCall _ | EExtern _
+  | EIf _ | ESelect _ | EBOp _ | EUOp _ | EList _ | EPub _ ->
+    tr_pattern ~public p, scheme_wildcard p.pos
```
ppolesiuk, line 356: The same problem as in line 370.

src/DblParser/Desugar.ml
```diff
@@ -376,13 +389,7 @@ let tr_param_decl (fld : Raw.field) =
...
     Either.Left (TNVar x, x, k)
   | FldName n ->
-    let sch =
-      { sch_pos   = fld.pos;
-        sch_args  = [];
-        sch_body  = make TWildcard
-      }
-    in
-    Either.Right (n, ident_of_name n, sch)
+    Either.Right (n, ident_of_name n, scheme_wildcard fld.pos)
```
ppolesiuk, line 392: Parameters without an annotation are always monomorphic, but maybe they should be treated uniformly with named patterns.

src/DblParser/Desugar.ml
```diff
@@ -401,16 +408,17 @@ let tr_named_arg (fld : Raw.field) =
   | FldTypeVal(x, arg) ->
     make (NP_Type(false, make (TNVar x, tr_type_arg arg)))
   | FldName n ->
-    make (NP_Val(n, make (PId(false, ident_of_name n))))
+    make (NP_Val(n, make (PId(false, ident_of_name n)),
+                 scheme_wildcard fld.pos))
```
ppolesiuk, line 412: Is this function the same as `tr_named_pattern ~public:false`? Maybe we can simplify desugar further.

ppolesiuk (changes requested):
I think, these changes fix one bug, but introduce an another. See comments below.

src/DblParser/Desugar.ml
forell: They're very similar, but currently `tr_named_pattern` rejects anonymous types at parsing time, while `tr_named_arg` (obviously) permits them. I think it'd be better if `tr_named_pattern` accepted those type binders and `tr_named_arg` was deleted, though, since the type checker can deal with cases when anonymous type binders don't make sense. Should I make this change?

ppolesiuk: Yes, I think it would be nice to have desugaring simpler. `TypeInference.Pattern.check_named_pattern` raises an error on anonymous type patterns, so there is no need to raise this error during parsing/desugaring.

forell:
Hopefully a bit less buggy now. I got rid of `tr_named_arg` and made `parameter` scheme annotations optional in `Surface`, too.

PR #182, opened 2025-02-18
Additions mutable lib

Kiczor:
Added some basic functions on arrays to Mutable library: map, concatenate, foldLeft, foldRight, exists, rev

### DIFF
--- a/lib/Mutable.fram
+++ b/lib/Mutable.fram
@@ -246,9 +246,9 @@ pub method iteri {E, self : Array E _} f =
   loop 0
 
 {## Perform right fold on array.
-  `a.fold_right f init [a1, a2, ..., an]` is equivalent to `f a1 (f a2 (... (f an init) ...))` 
+  `a.foldRight f init [a1, a2, ..., an]` is equivalent to `f a1 (f a2 (... (f an init) ...))` 
 ##}
-pub method fold_right {E, self : Array E _} f init =
+pub method foldRight {E, self : Array E _} f init =
   let rec loop (i : Int) =
     if i >= self.length then init
     else f (unsafeGetArray {E} self i) (loop (i + 1))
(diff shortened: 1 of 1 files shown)


lib/Mutable.fram
```diff
@@ -251,6 +251,80 @@ pub method iteri {E, self : Array E _} f =
...
+      self.at i := (f (unsafeGetArray {E} self i));
+      loop (i + 1))
+  in
+  loop 0
+
+{## Concatenate two arrays.
+  `a.concatenate b` returns new array with elements from `a` followed by elements from `b`
+##}
+pub method concatenate {E, self : Array E _} (other : Array E _) =
+  let result = unsafeMakeArray {E} (self.length + other.length)
```
Foxinio, line 291: In my opinion this and any function using `unsafeMakeArray` should use `unsafeInitArray` instead.

lib/Mutable.fram
```diff
@@ -251,6 +251,80 @@ pub method iteri {E, self : Array E _} f =
...
+  let rec loop (i: Int) acc = 
+    if i >= self.length then acc
+    else loop (i + 1) (f acc (unsafeGetArray {E} self i))
+  in
+  loop 0 init
+
+{## Map a function over an array.
+ `arr.map f` maps function `f` over array `arr`
+##}
+pub method map {E, self : Array E _} f =
```
Foxinio, line 278: Taking in any function to method working on Arrays should be written carefully, because of possible effects and resumptions. While this may be done correctly, not in place version of map may need to place some restrictions on effects of `f`>

lib/Mutable.fram
```diff
@@ -251,6 +251,80 @@ pub method iteri {E, self : Array E _} f =
...
+      loopset (i + 1) )
+  in
+  loopset 0;
+  result
+
+{## Check if any element of an array satisfy a condition.
+  `a.exists f` returns `True` if function `f` returns `True` on any element of an array `a`,
+   otherwise it returns `False`
+##}
+pub method exists {E, self : Array E _} f =
```
Foxinio, line 321: When we have `exists`, it will be good to have `all` or `forall`.

Foxinio (changes requested):
Aside from some minor suggestions, these changes look very good. However I would consider adding both versions in place and the other one, for mutable arrays.

KR338:
Going to copy this branch and finish this as agreed during today's meeting

PR #183, opened 2025-02-26
Operations on scopes

ppolesiuk:
- Renamed `root` to `initial` scope.
- Added `root` and `any` scopes. Root should not contain any type variables and is a parent of `initial`. This approach will be convenient in constraint solving: final solving of constraints will be just fixing scope of constraints to `root` and setting all remaining unification variables to pure.
- Added `parent`, `equal`, `subset`, and `strict_subset` operations. `subset` is basically the same as `mem`, but is used for different purposes, so more readable name will be convenient.
- `mem` function is not used directly, but wrapped around `TVar.in_scope` and `UVar.in_scope` functions.

### DIFF
--- a/src/Lang/Unif.mli
+++ b/src/Lang/Unif.mli
@@ -556,6 +556,9 @@ module TVar : sig
   (** Get the scope of a type variable *)
   val scope : tvar -> Scope.t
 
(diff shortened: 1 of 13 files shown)


forell (approved):
Looks good.

PR #187, opened 2025-03-17
Core language prepared for the new effect inference

ppolesiuk:
- Added effect constraints in Core.
- Modified definition of subeffecting in Core in a way that takes into account effect constraints.
- Modified the notion of productivity in recursive definitions: mutually recursive definitions are pure expressions (not necessarily values) that may refer to themself only under (but not necessarily directly) a new `ERecCtx` construct. This new construct is always impure.
- Modified Eval and Untyped, to allow recursive definitions to be any expression (but they must be productive, and use their continuation in a linear way).

### DIFF
--- a/src/Eval/Env.ml
+++ b/src/Eval/Env.ml
@@ -4,16 +4,25 @@
 
 open Value
 
-type t = value Var.Map.t ref
+type value_box = value option ref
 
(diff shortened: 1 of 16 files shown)


forell (approved):
Looks ok, merging.

PR #189, opened 2025-03-25
Achieving compatibility with ocaml 5.3

Foxinio:
I changed every occurence of `effect` to `effct` to satisfy ocaml 5.3 new `effect` keyword.
It seems like after that project compiles and all tests pass.


forell (approved):
Thanks!

PR #190, opened 2025-03-28
ConE intermediate language and simple incremental SAT-solver

ppolesiuk:
This pull request introduces ConE intermediate language ­— the internal language of effect inference. In ConE, effects can be guarded by (positive) formulae of propositional logic, and from these formulae we generate clauses for a SAT-solver. For convenience, ConE shares the representation of propositional variables with the SAT-solver, which is also included in this PR.


forell (approved):
Looks good. The implementation of the SAT solver is not really documented, so if we want to keep it around for a while then some comments describing the overall strategy and the role of some of the functions would help.

PR #196, opened 2025-04-01
Fixed bug in backtrackable references

ppolesiuk:
Fixed bug in backtrackable references, that allowed changes from committed world to be visible in the active world, even if there is an invalid world in between. This commit fixes wrong behavior of REPL when user-defined handlers were installed during the REPL session.

Additionally, I added some `to_sexpr` functions to SAT-solver, that I implemented during the debugging process. I hope they might be useful in the future.

### DIFF
--- a/src/IncrSAT/Solver.ml
+++ b/src/IncrSAT/Solver.ml
@@ -217,3 +217,15 @@ let solve_all solver =
       end
     end
   | err -> err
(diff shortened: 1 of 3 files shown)


dominik-muc (approved):

Foxinio:
LGTM

PR #213, opened 2025-04-02
update pythagorean triples examples to support new syntax

dominik-muc:


wojpok:
I like code improvements. Comments on the other hand are too excessive to my liking; for example there is no need to write such an elaborate explanation about what a Triples of Ints are.

dominik-muc:
This is an example — thus it's probably gonna be read by someone who is a beginner. The comment notices that this construction is actually represented as an ADT, which might not be so obvious.

wojpok:
Arguably, this is the least beginner friendly example. The third definition introduces effects and explaining it in terms of continuation doesn't clarify much to beginners. Without experience in functional programming this is as unreadable as it might be.

Speaking of effects I have just noticed a error in explanation of BT:
```
Fail, on the other hand, returns some value of a type T.
T is a local existential type, which means that each instance of BT
will implement this with a specific type.
```
In `fail : Unit ->[E] T` type `T` is not really an existential type, value is never instantiated and never returned. On `fail`, both handlers discards continuation, which means that `fail` doesn't return and thus might be used in any context.

ppolesiuk:
Now, we have a nice tutorial, so detailed explanations in the example are not needed anymore.

PR #214, opened 2025-04-02
Type-level infix operators

wbukowski:


wbukowski:
Should I reintroduce some of the types (e.g. `Pair`, `Either`) as operators as part of this pull request?

Foxinio:
> Should I reintroduce some of the types (e.g. `Pair`, `Either`) as operators as part of this pull request?

That's a good idea. We will immediately have some usage that will in small way test your changes.

ppolesiuk (changes requested):
Most of the changes looks good. However, there are two things that need to be fixed:
- there is no possibility of accessing type-level operator from a module, e.g. `M.[+]` (if this will be fixed, the `TBOpID` constructor should not be necessary, `TVar` will be sufficient)
- we agreed that allowing definitions like `data A + B = ...` instead of `data [+] A B = ...` are misleading: analogous construct on the level of expressions means something completely different (pattern matching against the constructor-level operator). This should be disallowed, at least by some warning.

rureirureirurei:
+

PR #215, opened 2025-04-03
add: syntactic sugar for effectful functions

### LINKED ISSUE #177: Syntactic sugar for effectful functions
Because effectful functions, where effects are inferred by the type-checked are quite common, we agreed on a syntactic sugar for them. Arrow `->>` should be a syntactic sugar for `->[_]`.

dominik-muc:
Closes #177


src/DblParser/YaccParser.mly
```diff
@@ -11,7 +11,7 @@
 %token<string> STR
 %token<char> CHR
 %token BR_OPN BR_CLS SBR_OPN SBR_CLS CBR_OPN CBR_CLS
-%token ARROW ARROW2 BAR COLON COMMA DOT EQ SEMICOLON2 SLASH GT_DOT
+%token ARROW ARROWW ARROW2 BAR COLON COMMA DOT EQ SEMICOLON2 SLASH GT_DOT
```
Foxinio, line 14: This is minor, but i think this deserves a better name, because in few weeks no one will know what is `ARROWW` without looking to the lexer. I propose `EFFECTFUL_ARROW` or `EFF_ARROW`.

src/DblParser/YaccParser.mly
```diff
@@ -145,6 +145,8 @@ op
 
 ty_expr
 : ty_expr_app ARROW ty_expr { make (TArrow($1, $3)) }
+| ty_expr_app EFF_ARROW ty_expr
+    { make (TArrow($1, make (TApp(make (TEffect [make TWildcard]), $3)))) }
```
ppolesiuk, line 149: Effect arrow is desugared here, in the parser, not the big `Desugar` module. Maybe it is right decision, because `Desugar` is too big, and too complicated, but I would like to make such decisions consciously.

But if we desugar it here, It would be nice to have more precise information about locations. Here, the location of the effect expression is the same as the whole arrow expression. Maybe it would be better to point to only the arrow.

PR #217, opened 2025-04-04
Printing location of unsolved unification variables

Foxinio:
Before getting `Unsolved unification variable left` meant walking in the dark trying to figure, where this error came from.
Now when seeing this error it will be accompanied by location where the unification variable was created:
```
fatal error: Unsolved unification variable left
 -> test1.fram
 1 | let id a = a
 2 |>let _ = id
   |         ^^
```

### DIFF
--- a/src/EffectInference/Error.ml
+++ b/src/EffectInference/Error.ml
@@ -129,6 +129,4 @@ let unused_pattern ~pos =
   (pos, "This pattern is unused")
 
 let unsolved_unification_variable ~pos =
-  InterpLib.Error.report ~pos ~cls:FatalError
-    "Unsolved unification variable left";
-  raise InterpLib.Error.Fatal_error
+  (pos, "Unsolved unification variable left")
--- a/src/EffectInference/Error.ml
+++ b/src/EffectInference/Error.ml
@@ -128,7 +128,7 @@ let non_exhaustive_match ~pos ctx =
 let unused_pattern ~pos =
   (pos, "This pattern is unused")
 
-let unsolved_unification_variable () =
-  InterpLib.Error.report ~cls:FatalError
+let unsolved_unification_variable ~pos =
+  InterpLib.Error.report ~pos ~cls:FatalError
     "Unsolved unification variable left";
   raise InterpLib.Error.Fatal_error
--- a/src/TypeInference/Expr.ml
+++ b/src/TypeInference/Expr.ml
@@ -209,8 +209,6 @@ let rec check_repl_def_seq ~tcfix env def_seq tp =
           match tp_req with
           | Check tp -> (tp, Checked)
           | Infer    ->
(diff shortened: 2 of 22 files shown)


src/EffectInference/Error.ml
```diff
@@ -128,7 +128,7 @@ let non_exhaustive_match ~pos ctx =
 let unused_pattern ~pos =
   (pos, "This pattern is unused")
 
-let unsolved_unification_variable () =
-  InterpLib.Error.report ~cls:FatalError
+let unsolved_unification_variable ~pos =
+  InterpLib.Error.report ~pos ~cls:FatalError
```
ppolesiuk, line 132: With position attached, this error can be treated as other errors in this module: they can be translated to a pair position-message, and raised using `fatal` function.

src/TypeInference/Expr.ml
```diff
@@ -209,7 +209,9 @@ let rec check_repl_def_seq ~tcfix env def_seq tp =
           match tp_req with
           | Check tp -> (tp, Checked)
           | Infer    ->
-            let tp = Env.fresh_uvar env T.Kind.k_type in
+            (* This probably can be something more than nowhere,
+                but will require significant refactor *)
```
ppolesiuk, line 213: I agree, and I think that we should just don't care about that. Showing locations of unsolved unification variables would make sense in batch mode, but in REPL we are going to tolerate such variables. In the future it will be impossible to see an error at this location.

src/Utils/Position.ml
```diff
@@ -75,6 +75,39 @@ let join p1 p2 =
   ; pos_end_bol    = p2.pos_end_bol
   }
 
+let join_sorted p1 p2 =
```
ppolesiuk, line 78: This function is not documented.

src/Utils/Position.ml
```diff
@@ -75,6 +75,39 @@ let join p1 p2 =
...
+  and pos_length = abs (p2.pos_start_cnum - p1.pos_start_cnum)
+    + max p1.pos_length p2.pos_length
+  and pos_fname = p1.pos_fname
+  in { pos_fname; pos_start_line; pos_start_cnum
+  ; pos_start_bol; pos_length; pos_end_line; pos_end_bol }
+
+(** Join list of positions into single one.
+  It take file name the head of the list and join the rest.
+  Returns Position.nowhere for empty list *)
+let join_list f xs =
```
ppolesiuk, line 104: Where this function is used? Do we need it?

ppolesiuk (commented):
Thanks! The changes look good and enhance user experience. I opt for merging them, but first some minor things should be improved. See comments below.

wojpok (approved):
Cool feature, looks good

dominik-muc (approved):

PR #218, opened 2025-04-04
fix: allow positively recursive types

### LINKED ISSUE #198: Non consistent approach to positive recursion
The type-checker allows pure pattern-matching on positively recursive types, while internal Core type-checker uses strict positivity requirement. This discrepancy may result in internal errors in some rare cases. For instance, for the program
```
data rec T = C of ((T -> Unit) -> Unit)
let foo (C f) = f (fn (C g) => g (fn _ => ()))
```
we get the following error.
```
Fatal error: exception Failure("Internal error: Type is not strictly positively recursive")
```

dominik-muc:
I'm not sure it's the correct approach for checking positive recursion, please verify me.


src/Lang/CorePriv/Syntax.ml
```diff
@@ -14,7 +14,7 @@ type data_def =
       proof             : var;
       args              : TVar.ex list;
       ctors             : ctor_type list;
-      strictly_positive : bool
+      positive : bool
```
ppolesiuk, line 17: Code formatting could be improved here. All colons above are aligned, so why to do it differently.

src/Lang/CorePriv/Type.ml
```diff
@@ -418,41 +418,47 @@ let rec strictly_positive : type k. nonrec_scope:_ -> k typ -> bool =
     end
 
   | TEffJoin(eff1, eff2) ->
-    strictly_positive ~nonrec_scope eff1 &&
-    strictly_positive ~nonrec_scope eff2
-
+    positive ~nonrec_scope eff1 &&
+    positive ~nonrec_scope eff2
+
+  | TArrow(TArrow (_, tp1, eff0), tp2, eff1) ->
```
ppolesiuk, line 424: Ugh... Onion is either nothing or an onion wrapped around a leaf. Most of operations on syntax should deconstruct at most one constructor at once. I think, we need two mutually recursive functions here: `positive` and `negative`.

src/ToCore/DataType.ml
```diff
@@ -73,7 +73,7 @@ let finalize_data_def env (dd : data_def) =
       proof             = dd.proof;
       args              = args;
       ctors             = ctors;
-      strictly_positive = dd.positive;
+      positive = dd.positive;
```
ppolesiuk, line 76: Code formatting again...

test/ok/ok0129_positiveAdt.fram
```diff
@@ -0,0 +1,2 @@
+data rec T = C of ((T -> Unit) -> Unit)
+let foo (C f) = f (fn (C g) => g (fn _ => ()))
```
ppolesiuk, line 2: We are not very strict about that (unfortunately), but this is not a valid Unix text file: it has no LF at the end.

ppolesiuk (changes requested):
Thanks for this PR. Unfortunately, the positivity checking in Core is not implemented correctly (and is bit inelegant). This should be fixed before merging. See comments below.

src/Lang/CorePriv/Type.ml
dominik-muc: I'm sorry, I don't follow. Isn't negative just *not positive*?

ppolesiuk: No. For instance, parameter of application isn't on positive nor negative position. When some type is raised by subtyping,
types/effects on positive positions can raised too, while types/effects on negative positions can be lowered.

dominik-muc:
Tests failed on opam setup... Can you re-run the CI?

src/Lang/CorePriv/Type.ml
```diff
@@ -418,41 +418,69 @@ let rec strictly_positive : type k. nonrec_scope:_ -> k typ -> bool =
...
+    | true, Some _ -> true
     | _ -> false
     end
 
+(** Check if all types on non-negative positions fits in given
+  scope. *)
+and negative : type k. nonrec_scope:_ -> k typ -> bool =
+  fun ~nonrec_scope tp ->
+  match tp with
+  | TVar _ | TEffPure -> true
```
ppolesiuk, line 446: The type variable is on positive, and therefore, on non-negative position, so we have to check if it fits in the given scope. It can be done the same way as constructs in the next line, i.e, `TLabel _ | TData _`.

src/Lang/CorePriv/Type.ml
```diff
@@ -418,41 +418,69 @@ let rec strictly_positive : type k. nonrec_scope:_ -> k typ -> bool =
...
+  | TGuard _ | TLabel _ | TData _ ->
+    begin match type_in_scope nonrec_scope tp with
+    | Some _ -> true
+    | None   -> false
+    end
+
+  | TEffJoin(eff1, eff2) ->
+    negative ~nonrec_scope eff1 &&
+    negative ~nonrec_scope eff2
```
ppolesiuk, line 455: All occurrences of variables in effects are positive, and therefor non-negative. This case can be merged with the previous one.

src/Lang/CorePriv/Type.ml
```diff
@@ -418,41 +418,69 @@ let rec strictly_positive : type k. nonrec_scope:_ -> k typ -> bool =
...
     | _ -> false
     end
 
+(** Check if all types on non-negative positions fits in given
+  scope. *)
+and negative : type k. nonrec_scope:_ -> k typ -> bool =
+  fun ~nonrec_scope tp ->
+  match tp with
+  | TVar _ | TEffPure -> true
+  | TGuard _ | TLabel _ | TData _ ->
```
ppolesiuk, line 447: We can do better with `TGuard` construct. Body of the guard is on the positive position. Same for the left-hand-side of constraints. Right-hand-side is on negative position. We can do the same in `positive` function.

ppolesiuk (changes requested):
Now, it looks better, but still needs to be improved. As for now, functions `positive` and `negative` are equivalent, but they should be different. See comments below.

src/Lang/CorePriv/Type.ml
```diff
@@ -405,54 +405,79 @@ and subtype_in_scope scope (tp : ttype) =
...
     | None   -> false
     end
+  
+  | TGuard (constraints, tp) ->
+    positive ~nonrec_scope tp &&
+    List.for_all (fun (lhs, rhs) ->
+      positive ~nonrec_scope lhs &&
+      negative ~nonrec_scope rhs
+    )
+    constraints
```
ppolesiuk, line 426: Code formatting might be improved, to make it more similar to existing code base. The closing parenthesis should not be in a separate line. It might look like:
```
List.for_all
  (fun ... ->
    ...)
  constraints
```
or
```
constraints |> List.for_all
  (fun ... ->
    ...)
```

src/Lang/CorePriv/Type.ml
```diff
@@ -405,54 +405,79 @@ and subtype_in_scope scope (tp : ttype) =
...
+  fun ~nonrec_scope tp ->
+  match tp with
+  | TEffPure -> true
+  | TVar _  | TLabel _ | TData _ | TApp _ | TEffJoin _ ->
+    begin match type_in_scope nonrec_scope tp with
+    | Some _ -> true
+    | None   -> false
+    end
+
+  | TGuard _  -> positive ~nonrec_scope tp
```
ppolesiuk, line 461: It doesn't seem right... Something is negative when is positive?!

ppolesiuk (changes requested):
I think, the `TGuard _` case in `negative` should be fixed. Also, code formatting might be improved.

src/Lang/CorePriv/Type.ml
dominik-muc: I misunderstood when you said "We can do the same in `positive` function."...

src/Lang/CorePriv/Type.ml
```diff
@@ -405,54 +405,85 @@ and subtype_in_scope scope (tp : ttype) =
...
     | None   -> false
     end
+  
+  | TGuard (constraints, tp) ->
+    positive ~nonrec_scope tp &&
+    List.for_all
+      (fun (lhs, rhs) ->
+        positive ~nonrec_scope lhs &&
+        negative ~nonrec_scope rhs)
+    constraints
```
ppolesiuk, line 426: In this line, there is a parameter to function `List.for_all`. It should be indented, as other parameters, i.e, lambda. It should be:
```
List.for_all
  (fun (lhs, rhs) ->
    positive ~nonrec_scope lhs &&
    negative ~nonrec_scope rhs)
  constraints
```

src/Lang/CorePriv/Type.ml
```diff
@@ -405,54 +405,85 @@ and subtype_in_scope scope (tp : ttype) =
...
 
   | TEffJoin(eff1, eff2) ->
-    strictly_positive ~nonrec_scope eff1 &&
-    strictly_positive ~nonrec_scope eff2
-
+    positive ~nonrec_scope eff1 &&
+    positive ~nonrec_scope eff2
+
```
ppolesiuk, line 431: This spaces in the empty line might be removed.

src/Lang/CorePriv/Type.ml
```diff
@@ -405,54 +405,85 @@ and subtype_in_scope scope (tp : ttype) =
...
+    | None   -> false
+    end
+
+  | TGuard (constraints, tp) ->
+    negative ~nonrec_scope tp &&
+    List.for_all
+      (fun (lhs, rhs) ->
+        negative ~nonrec_scope lhs &&
+        positive ~nonrec_scope rhs)
+    constraints
```
ppolesiuk, line 467: Same comment as in `positive` function, i.e., more indentation.

ppolesiuk (approved):
Looks good, only minor things regarding formatting might be improved. Thanks!

ppolesiuk:
@forell Could you review this PR? It touches important parts of the Core, so I want to make sure I didn't overlooked something.

PR #219, opened 2025-04-08
fix: bounds check in Int64 to Int conversion

dominik-muc:


forell (changes requested):
The changes look good to me. I'd prefer to see the constant names in `/Base/Int` changed, though. Also, perhaps it'd be nice if the Prelude contained an `Int` module exporting those constants for convenience, now that they're available.

forell (approved):
Looks good, thanks!

PR #237, opened 2025-05-01
Check for ungeneralizable parameters in pattern lets

forell:
A call to `end_generalize_impure` was missing for let expressions with patterns. As a result, programs like the following failed with a much less informative internal error.
```fram
parameter x : Unit
let () = x
```


ppolesiuk (approved):
Thanks for finding this bug! Approve.

PR #242, opened 2025-05-18
Type synonyms without parameters

### LINKED ISSUE #223: Type synonyms without parameters
To implement type synonyms I would start with the simplest variant, when type synonyms cannot have any parameters.

ppolesiuk:
This PR introduces type synonyms in the simplest form: when they have no parameters. However, this PR setups ground for further extensions related to type synonyms.

Resolves #223


wojpok (approved):

forell (approved):
I'm in favor of merging this PR.

As a related comment, while the synonyms work nicely for grouping several effects, they're useless for pretty-printing them. I can't think of any simple solution to make this work. Should this be added as a new issue?
```fram
> type E = [ST, IO];;
> let f (g : Int ->[E] Int) = g 42;;
> f;;
: (Int ->[IO,ST] Int) ->[IO,ST,a] Int
```

src/TypeInference/Unification.ml
```diff
@@ -69,14 +69,19 @@ let kind_to_arrow k =
 
   | KType | KEffect -> None
 
-let set_uvar env u tp =
+let rec set_uvar env u tp =
   if T.Type.contains_uvar u tp then
-    raise Error
+    begin match T.Type.view tp with
+    | TUVar u2 when T.UVar.equal u u2 -> ()
+    | TAlias(_, tp) -> set_uvar env u tp
```
Foxinio, line 76: As discussed during meeting, this should has comment attached, describing this edge case.

ppolesiuk: I've added a comment that explains the edge case.

Issue #220, opened 2025-04-12
Special implicits with source location


ppolesiuk:
We could add implicit parameters that are resolved in a special way, i.e., they are automatically instantiated with current filename, line, or other metadata about source location. Their names should start with `__` (for instance `~__line__`) to signify that they have special meaning. Then we can use them in `runtimeError`, `assert`, and `impossible` functions of the standard library to enhance error messages.

This is rather easy task. Only the `TypeInference.ParamResolve.resolve_implicit` function should be modified.

PR #243, opened 2025-05-23
Line and File implicits

wojpok:
This PR introduces `~__line__` and `~__file__` implicit parameters which are instantiated with line number and file name respectively, unless defined otherwise. This allows to enhance `assert` function.

Solves #220.

### DIFF
--- a/src/TypeInference/ParamResolve.ml
+++ b/src/TypeInference/ParamResolve.ml
@@ -219,12 +219,28 @@ and resolve_optional ~vset ~pos env rctx x sch =
 and resolve_implicit ~vset ~pos env rctx iname sch =
   let name = Name.NImplicit iname in
   match ModulePath.try_lookup_implicit ~pos env iname with
-  | None ->
-    Error.fatal (Error.cannot_resolve_implicit ~pos iname)
   | Some(x, x_sch) ->
     let vset = restrict_var ~vset ~pos ~pp:(Env.pp_tree env) x name in
     let e = { T.pos; T.pp = Env.pp_tree env; T.data = T.EVar x } in
     coerce_scheme ~vset ~pos ~name env e x_sch sch
+  | None ->
+    (* Special implicits *)
+    let pp = Env.pp_tree env in
+    let make data = T.{data; pos; pp} in
+    let (param, scheme) = match iname with
+      | "~__line__" -> 
+        (make (T.ENum pos.pos_start_line), BuiltinTypes.int_scheme)
+      | "~__file__" -> 
+        (make (T.EStr pos.pos_fname), BuiltinTypes.string_scheme)
+      | _ -> Error.fatal (Error.cannot_resolve_implicit ~pos iname) in
+    (* Check types *)
+    let (env, _, _, _, tp_out) = open_scheme ~pos env sch in
+    let (sub, _) = guess_types ~pos env scheme.sch_targs in
+    let tp_in = T.Type.subst sub scheme.sch_body in
+    Error.check_unify_result ~pos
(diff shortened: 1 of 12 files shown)


Foxinio (commented):
Idea is great and changes look good. As a one suggestion, there are many stdlib functions (like method div) that use assertions and once they fail it probably would be better to see file and line number, of the place div was called, not the place div is implemented. As far as I understand the implementation, to make this work every function or method making any assertions would need to bind an implicit, and pass it to the assertion.
I'm not saying it should be addressed in this PR, however if it is not, an issue should be created describing the improvement.

wojpok:
Good point. We definitely have to decide how deep we want to propagate those implicits, because once captured, they cant be reseted. It might be a valid argument to introduce line and file optionals. There even might be a way to capture a part of a stacktrace.

src/TypeInference/ParamResolve.ml
```diff
@@ -219,12 +219,28 @@ and resolve_optional ~vset ~pos env rctx x sch =
...
+    let (param, scheme) = match iname with
+      | "~__line__" -> 
+        (make (T.ENum pos.pos_start_line), BuiltinTypes.int_scheme)
+      | "~__file__" -> 
+        (make (T.EStr pos.pos_fname), BuiltinTypes.string_scheme)
+      | _ -> Error.fatal (Error.cannot_resolve_implicit ~pos iname) in
+    (* Check types *)
+    let (env, _, _, _, tp_out) = open_scheme ~pos env sch in
+    let (sub, _) = guess_types ~pos env scheme.sch_targs in
+    let tp_in = T.Type.subst sub scheme.sch_body in
```
ppolesiuk, line 239: I think this piece of code is overcomplicated. All these special implicits are monomorphic, and I don't see any scenario where we could introduce polymorphic special implicits to the language. This code can be simplified.

Btw, is there any reason to not use `coerce scheme` function, like in the case of regular implicit?

ppolesiuk (commented):
The changes looks good, and generally I would like to approve them. However, the main implementation of the feature seems to be complicated, as described in the comment below.

src/TypeInference/ParamResolve.ml
```diff
@@ -219,12 +219,28 @@ and resolve_optional ~vset ~pos env rctx x sch =
...
+    Error.check_unify_result ~pos
+      (Unification.subtype env tp_in tp_out)
+      ~on_error:(Error.named_param_type_mismatch ~pp name tp_in tp_out);
+    (make (T.PF_Fun ([], [], param)), [])
```
ppolesiuk, line 243: A bit simpler implementation, that uses `coerce_scheme` function and takes advantage of the fact, that all special implicits have base type. This implementation doesn't need `BuiltinType.int_scheme` and `BuiltintType.string_scheme`, so they can be removed.

```suggestion
    let (param, scheme) = match iname with
      | "~__line__" -> 
        (make (T.ENum pos.pos_start_line), BuiltinTypes.int_scheme)
      | "~__file__" -> 
        (make (T.EStr pos.pos_fname), BuiltinTypes.string_scheme)
      | _ -> Error.fatal (Error.cannot_resolve_implicit ~pos iname) in
    (* Check types *)
    let (env, _, _, _, tp_out) = open_scheme ~pos env sch in
    let (sub, _) = guess_types ~pos env scheme.sch_targs in
    let tp_in = T.Type.subst sub scheme.sch_body in
    Error.check_unify_result ~pos
      (Unification.subtype env tp_in tp_out)
      ~on_error:(Error.named_param_type_mismatch ~pp name tp_in tp_out);
    (make (T.PF_Fun ([], [], param)), [])
```

PR #247, opened 2025-06-02
Recurse correctly on ERecCtx

### LINKED ISSUE #246: Infinite loop in Core dumping
Running 
```
let rec factorial (n : Int) = 
    if n == 0 then 1 
    else n * factorial (n - 1)
```
with `-dcore` results in an infinite loop in `CorePriv.SExprPrinter.tr_defs`.

Brychlikov:
Fixes #246.
Fixes a bug where `CorePriv/SExprPrinter.tr_defs` would not recurse properly on `ERecCtx`, instead falling into an infinite loop.

The question is, should this be tested? Currently, we don't test `-dcore` and `-dcone` flags anywhere in the suite, so I didn't add any tests confirming my fix.


dominik-muc (approved):
I don't believe we have tools necessary to test such thing. Nor that we should, it's an overlook mistake and a straight-forward fix.

PR #248, opened 2025-06-02
Union find

dominik-muc:
My first take on union find, what should I implement now to „connect” it to the ConstrSolve machinery? Detecting a cycle is easy -> when adding an edge (an union between two points) if they already were on the same component, a cycle is found (I understand this now :)). I can implement that for integers, but I don't know how should I represent cycles of effects. After detecting a cycle do I need to delete those nodes from the graph? That could be tricky. Do I first add all nodes and then search for cycles? That wouldn't make use of union-find structure. If someone is willing to look at that any advice would be nice, I'll try to improve it in the next days.

### DIFF
--- a/src/Utils/UnionFind.ml
+++ b/src/Utils/UnionFind.ml
@@ -1,0 +1,62 @@
+
+module type OrderedType =
+  sig
+    type t
+    val compare: t -> t -> int
+  end
+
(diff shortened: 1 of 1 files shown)


dominik-muc:
Ahh the problem is that we have a *directed* graph. It's over

Issue #19, opened 2024-03-23
REPL pretty-printing based on `toString` method


ppolesiuk:
The current pretty-printer in REPL behaves very poorly.
```
$ dbl
> data T = A ;;
> A ;;
: T
= <ctor>
```
Better pretty-printer could base on `toString` method.

PR #249, opened 2025-06-14
Implementation of pretty printing using toString method.

Dominik-Pawelec:
Implementation of #19 - REPL pretty printing using toString method defined in fram.
Added additional flag -no-toString-printing which sets printing to default implementation.

### DIFF
--- a/src/ConETypeErase.ml
+++ b/src/ConETypeErase.ml
@@ -65,8 +65,10 @@ let rec tr_expr (e : S.expr) =
   | ERepl(func, _, _) ->
     T.ERepl (fun () -> tr_expr (func ()))
 
-  | EReplExpr(e1, tp, e2) ->
-    T.EReplExpr(tr_expr e1, tp, tr_expr e2)
+  | EReplExpr(e1, tp, e2, ep) ->
+    match ep with
+    | None    -> T.EReplExpr(tr_expr e1, tp, tr_expr e2, None)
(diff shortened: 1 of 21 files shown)


ppolesiuk (changes requested):
Thanks for sending PR. Overall, this looks like a solid piece of work. However, some minor issues should be addressed. Most importantly, I concern about the data representation in ConE, Core, and Untyped. For me it contains redundant information: if the `toString` method is used, the `EReplExpr` constructor still keeps the original expression, but it is never used. I would rather prefer to keep only one expression (wrapped around call to `toString` or not), and the flag indicating which method of printing was used. On the other hand, the chosen representation in Unif is can be defended: during the effect inference we still may want to fallback to default implementation (can we make this decision during type-inference?), and we need to infer and pretty-print the type of the original expression.

ppolesiuk (approved):
For me, the PR is ready for merge. However, I changed most of the code, so someone else should look at it.

src/Eval/External.ml
```diff
@@ -12,6 +12,7 @@ let extern_map =
                               str_fun (fun msg -> 
                                 runtime_error_with_postion fname line msg)));
       "dbl_magic", pure_fun Fun.id;
+      "dbl_toString",    pure_fun (fun v -> VStr (Value.to_string v));
```
forell: Alignment whitespace that doesn't seem to align `pure_fun ...` with anything.

PR #255, opened 2025-08-02
Simplified type annotations in Prelude format methods

wojpok:
Thanks to recent fixes to parameters/method resolve, format method in Prelude can be simplified by using universally defined `parameter method format` for X and Y


ppolesiuk:
After recent changes, this PR is outdated. Module `Base/Format` was renamed to `Base/Show` and already contains proposed changes. Therefore, we can close this PR now.

Issue #170, opened 2024-12-20
Expose maximum array length


ppolesiuk:
> In a discussion @ppolesiuk suggested that we could take an approach similar to OCaml and expose the maximum array length for when the programmer requires this kind of assurance. I think this is reasonable, given that oversized allocations are uncommon and recovery from out of memory situations is rarely needed/desirable (at least for such a high level language). However, perhaps we still want to check the size somewhere, to display a custom error instead of OCaml's.

_Originally posted by @forell in https://github.com/fram-lang/dbl/pull/126#discussion_r1652849679_

PR #257, opened 2025-10-22
Add: Expose maximum array length (#170)

kulson:
The developer can now write the following code:

```
$ dbl
> let maxLength = extern dbl_maxArrayLength : Int ;;
> maxLength ;;
: Int
= 18014398509481983
> 
```


wojpok:
Fine addition. Please, include this constant in Array module

PR #264, opened 2025-11-03
First-class lazy evaluation

wojpok:
This code is part of a pretty-printer PR #256 . This module have been separated from main PR for easier review and merging.

### DIFF
--- a/lib/Lazy.fram
+++ b/lib/Lazy.fram
@@ -2,10 +2,10 @@
  # See LICENSE for details.
  #}
 
-## # Fist-class lazy values
+## # First-class lazy values
 
 {##
-  This module provides first-class lazy values, also known assuspensions,
+  This module provides first-class lazy values, also known as suspensions,
   representing deferred computations evaluated on demand.
 
   A suspension encapsulates a computation that is executed only when its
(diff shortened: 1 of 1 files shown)


lib/Lazy.fram
```diff
@@ -0,0 +1,65 @@
...
+ ##}
+
+# Mutability utilities
+data RefP = RefP of {Ref : type -> type}
+
+let RefP {Ref} = (extern dbl_abstrType : Unit -> RefP) ()
+
+let ref {type X}    = (extern dbl_ref    : X -> Ref X)
+method get {type X} = (extern dbl_refGet : Ref X -> X)
+method set {type X} = (extern dbl_refSet : Ref X -> X -> Unit)
```
ppolesiuk: Operations on references are not pure. However, in this case, the effects of these operations cannot be observed directly, we cannot assume that these operations are very pure. Marking them we the empty non-pure effect would be a better choice.
```suggestion
let ref {type X}    = (extern dbl_ref    : X ->[] Ref X)
method get {type X} = (extern dbl_refGet : Ref X ->[] X)
method set {type X} = (extern dbl_refSet : Ref X -> X ->[] Unit)
```

PR #266, opened 2025-11-04
Streams

wojpok:
This PR is a part of Pretty-printer PR #256 . This module have been separated for easier review. Interface is not complete and might be expanded in future. This version provides minimal required set of utilities to run pretty-printer.

### DIFF
--- a/lib/Stream.fram
+++ b/lib/Stream.fram
@@ -8,7 +8,7 @@ import open /Lazy
 
 {##
   This module provides lazy lists, also known as streams.
-  Each and every node of a stream is deffered, meaniung that
+  Each and every node of a stream is deferred, meaning that
   no computation occurs unless results are forcibly read.
 
   Contrary to regular lists, streams may be infinite. Some iterative
@@ -31,15 +31,15 @@ method unstream (Stream xs) = xs
 pub let rec fromList (xs : List X) =
   Stream (lazy (fn _ =>
     match xs with
-    | x :: xs => Cons x (fromList xs)
     | []      => Nil
+    | x :: xs => Cons x (fromList xs)
     end)) : Stream X
 
 {## Returns all elements of a stream as a list. ##}
 pub let rec toList (Stream xs : Stream X) =
   match xs.force with
-  | Cons x xs => x :: toList xs
   | Nil       => [] : List X
+  | Cons x xs => x :: toList xs
   end : List X
 
 {##
@@ -55,10 +55,10 @@ pub let rec unfold {S} (seed : S) (f : S ->[] Option (Pair X S)) =
     end)) : Stream X
 
 {## Creates an empty stream. ##}
-pub let empty {X} = (Stream (pureLazy Nil) : Stream X)
+pub let empty = (Stream (pureLazy Nil) : Stream X)
 
 {## Checks if a stream is empty. ##}
-pub let isEmpty {X} (Stream xs : Stream X) =
+pub let isEmpty (Stream xs : Stream X) =
   match xs.force with
   | Nil => True
   | _   => False
@@ -72,7 +72,7 @@ pub let singleton (elem : X) =
 pub let cons (elem : X) (tail : Stream X) =
   Stream (pureLazy (Cons elem tail))
 
-{## Adds a deffered element to the head of a stream. ##}
+{## Adds a deferred element to the head of a stream. ##}
 pub let lazyCons (elem : Lazy X) (tail : Stream X) =
   Stream (lazy (fn _ => Cons elem.force tail))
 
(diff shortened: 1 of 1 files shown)


lib/Stream.fram
ppolesiuk: Since streams are parametrized by the type of the element, and almost all the functions in this module operates on streams of the same type (except `map` function), consider making type `X` a declared parameter. It would allow the programmer explicitly instantiate that type (when some missing type annotations would be added, e.g., in `fromList`).

ppolesiuk (commented):
The implementation looks solid. I have some minor comments regarding names of some functions, and documentation comments. A also would consider making type parameter `X` named for all functions through `parameter` mechanism.

lib/Stream.fram
```diff
@@ -0,0 +1,191 @@
...
+ ##}
+pub let rec init {X : type, Y : type}
+  (seed : X) (f : X ->[] Option (Pair Y X)) = (Stream (lazy (fn _ =>
+    match f seed with
+    | None             => Nil
+    | Some (val, seed) => Cons val (init seed f)
+    end)) : Stream Y)
+
+{## Creates an empty stream. ##}
+pub let mkEmpty {X : type} () = (Stream (pureLazy Nil) : Stream X)
```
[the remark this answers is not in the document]
wojpok: DBL is too restrictive for that to work. Importing Stream module in REPL with polymorphic empty value will throw "unresolved unification variable" error and simply refuse to load the module. It works fine when used in program where `empty` is used with concrete type at least once. Making it a function was the only way to make it load in any context. I can remove this function and use  `fromList` function instead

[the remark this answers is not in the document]
ppolesiuk: That's right, it's my mistake. `mkEmpty` sounds good to me.

lib/Stream.fram
ppolesiuk: While these comment was addressed in the recent changes, my actual suggestion was to use section mechanism: declare type `X` somewhere around the beginning of the file, like
```
parameter X
```
and use this `X` consistently to annotate *the* streams processed by subsequent functions.

Do you think, that it would increase the readability?

lib/Stream.fram
```diff
@@ -0,0 +1,199 @@
...
+end
+
+method unstream (Stream xs) = xs
+
+{## Creates stream from given list. ##}
+pub let rec fromList {X} (xs : List X) = Stream (lazy (fn _ =>
+  match xs with
+  | x :: xs => Cons x (fromList xs)
+  | []      => Nil
+  end)) : Stream X
```
ppolesiuk: I think, code formatting is inconsistent with the functions defined in the rest of the module. I would write something like the following.
```suggestion
pub let rec fromList {X} (xs : List X) =
  Stream (lazy (fn _ =>
    match xs with
    | x :: xs => Cons x (fromList xs)
    | []      => Nil
    end)) : Stream X
```

ppolesiuk (commented):
As you have asked, I wrote some comments about formatting. And there is still one unresolved conversation from my previous review.

lib/Stream.fram
wojpok: I am under Haskell influence and I tried to not break `Stream (lazy (fn =>` to separete line unless necessary. But I have to admit, the proposed style does look better

lib/Stream.fram
```diff
@@ -0,0 +1,199 @@
...
+      match (xs.force, ys.force) with
+      | Nil, Nil => True
+      | Cons x xs, Cons y ys =>
+        if x == y then
+          equalAux (xs : Stream X) ys
+        else
+          False
+      | _ => False
+    end in
+  equalAux xs ys
```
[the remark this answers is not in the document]
wojpok: I meant to align it with `let rec` definition above. My bad

lib/Stream.fram
```diff
@@ -0,0 +1,202 @@
...
+ ##}
+pub let rec unfold {S} (seed : S) (f : S ->[] Option (Pair X S)) =
+  Stream (lazy (fn _ =>
+    match f seed with
+    | None             => Nil
+    | Some (val, seed) => Cons val (unfold seed f)
+    end)) : Stream X
+
+{## Creates an empty stream. ##}
+pub let empty {X} = (Stream (pureLazy Nil) : Stream X)
```
ppolesiuk: Once `X` is declared as a section parameter, there is no need to bind it again.
```suggestion
pub let empty = (Stream (pureLazy Nil) : Stream X)
```

lib/Stream.fram
```diff
@@ -0,0 +1,202 @@
...
+      match (xs.force, ys.force) with
+      | Nil, Nil => True
+      | Cons x xs, Cons y ys =>
+        if x == y then
+          equalAux (xs : Stream X) ys
+        else
+          False
+      | _ => False
+    end in
+    equalAux xs ys
```
ppolesiuk: `end` keyword should be aligned with the matching `match`. Moreover, long time ago we had a discussion on formatting let definitions. A near consensus was to allow `in` keyword in a not separate line only for one-line definitions. With this convention, the code should look like that:
```suggestion
      end
    in
    equalAux xs ys
```

ppolesiuk (approved):
The PR is ready for merge. Thanks! The last comments from Copilot can be ignored, especially that on left to right direction of foldRight. It's a complete nonsense.

PR #268, opened 2025-11-05
Prelude extension

wojpok:
This PR is part of #256 separated for easier merging and review

### DIFF
--- a/lib/Prelude.fram
+++ b/lib/Prelude.fram
@@ -51,7 +51,7 @@ pub module Int64
   pub let ofInt (n : Int) = n.toInt64
 end
 
-## Returns the greater of two elements `x` and `y`. 
+## Returns the greater of two elements `x` and `y`.
 pub let max {T, method lt : T -> T ->> Bool} (x : T) (y : T) =
   if y < x then x else y
 


lib/Prelude.fram
```diff
@@ -48,3 +50,21 @@ pub module Int64
   pub let one  = 1L
   pub let ofInt (n : Int) = n.toInt64
 end
+
+pub let max (x : Int) (y : Int) =
+  if x > y then x else y
```
ppolesiuk: Is there any reason to define this function for `Int` type only? Maybe we can define a more polymorphic version for any type with `lt` method.
```suggestion
pub let max {T, method lt : T -> T ->> Bool} (x : T) (y : T) =
  if x < y then y else x
```
BTW, we could add also `min` function, and provide some documentation for them.

lib/Prelude.fram
```diff
@@ -48,3 +50,26 @@ pub module Int64
...
   pub let ofInt (n : Int) = n.toInt64
 end
+
+## Returns the greater of two elements `x` and `y`. 
+pub let max {T, method lt : T -> T ->> Bool} (x : T) (y : T) =
+  if y < x then x else y
+
+## Returns the smaller of two elements `x` and `y`.
+pub let min {T, method gt : T -> T ->> Bool} (x : T) (y : T) =
+  if x > y then y else x
```
ppolesiuk: Is there any reason to implement `min` and `max` functions on top of different comparison methods (`lt` and `gt`, respectively)? Both implementation can be based on `lt` method.
```suggestion
pub let min {T, method lt : T -> T ->> Bool} (x : T) (y : T) =
  if x < y then x else y
```

ppolesiuk:
```suggestion
pub let min {T, method gt : T -> T ->> Bool} (x : T) (y : T) =
  if y < x then y else x
```
?

wojpok: Yeah, right, my bad :sweat_smile:

wojpok:
Both `min` and `max` use `lt` method. Implementation have been supported with appropriate tests

Issue #232, opened 2025-04-15
Syntactic sugar for explicit instantiation with effectful functions


ppolesiuk:
Similarly to #228 we could have syntactic sugar for effectful functions, e.g.,
```
handle ~bt = BT
  { effect flip () / r = r True; r False,
  , effect fail () = ()
  }
```
This syntax should also work with other name kinds, like implicits (`{ effect ~n x = ... }`) or methods (`{ effect method m x = ... }`).

We also considered effects on explicit labels `{ l.effect x = ... }`, but we decided to not implement it, because this syntax is too complicated as for feature that is rarely used.

Issue #228, opened 2025-04-14
Syntactic sugar for explicit instantiations with functions


ppolesiuk:
Since named parameters are often instantiated with (polymorphic) lambda abstraction we could add convenient syntactic sugar for it. For instance, instead of
```
foo { bar = fn {T, x} y => ... }
```
we could write
```
foo { bar {T, x} y = ... }
```

PR #269, opened 2025-11-06
Syntactic sugar for explicit instantiations with (effectful) functions

wojciechdomin:
Added syntactic sugar for explicit instantiations with (effectful) functions, thus solving #228 and #232. 
Added 2 tests for this new feature.


src/DblParser/Desugar.ml
```diff
@@ -377,6 +377,7 @@ and tr_named_pattern ~public (fld : Raw.field) =
   | FldModule { data = NPName name; _ } -> make (NP_Module(public, name))
   | FldModule _ -> Error.fatal (Error.desugar_error fld.pos)
   | FldOpen     -> make (NP_Open public)
+  | FldNameFn _ | FldNameEffectFn _ -> assert false
```
ppolesiuk: This case is not impossible, for instance:
```
let (C { x y = z }) = 42
```
Some error should be provided for the user (for instance, you can use the default `desugar_error`).

ppolesiuk:
Is it still a draft? All issues seems to be resolved, so I think, you can change the status of this PR.

src/DblParser/YaccParser.mly
```diff
@@ -485,6 +484,9 @@ field
 | UID                   { make (FldType($1, None))    }
 | UID EQ ty_expr        { make (FldTypeVal($1, $3))   }
 | UID COLON kind_expr   { make (FldType($1, Some $3)) }
+| name expr_250_list1 EQ expr_no_comma { make (FldNameFn($1, $2, $4)) }
+| effect_label_opt KW_EFFECT name expr_250_list1 effect_resumption_opt EQ expr_no_comma
```
ppolesiuk: This line is too long.

src/DblParser/Raw.ml
```diff
@@ -78,6 +78,12 @@ type ('tp, 'e) field_data =
   | FldNameVal of name * 'e
     (** Named implicit parameter together with a value *)
 
+  | FldNameFn of name * 'e list * 'e 
+    (** Explicit instantiation with a function *)
+
+  | FldNameEffectFn of name * 'e option * 'e list * 'e option * 'e
+    (** Explicit instantiation with an effectful function *)
```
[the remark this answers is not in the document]
wojciechdomin: I disagree with this suggestion actually. It isn't consistent with the rest of comments.

ppolesiuk (approved):
Now, everything looks good. Thanks!

PR #278, opened 2025-11-20
Counter-examples of non-exhaustive pattern-matching

rureirureirurei:
How the error reporting used to be:
```
fatal error: This pattern-matching is not exhaustive.
 -> test.fram
...
```

Now, with this patch, unmatched pattern is printed

```
fatal error: This pattern-matching is not exhaustive.
Here is an example of a case that is not matched:
  Many (MyCons (Square) _)
  7 | let area = fn fig =>
  8 |>  match fig with
  9 |>  | Single Circle => 0
 10 |>  | Single Square => 0
 11 |>  | Single Triangle => 0
 12 |>  | Many Nil => 0
 13 |>  | Many (MyCons Circle other) => 0
 14 |>  end
```

### DIFF
--- a/src/EffectInference/PatternContext.ml
+++ b/src/EffectInference/PatternContext.ml
@@ -112,7 +112,12 @@ let pp_name (name : T.name) =
 (** Pretty-print an example pattern, wrapping in parens if needed *)
 let rec pp_ex_pattern_wrap need_parens ex =
   let pp = pp_ex_pattern ex in
-  if need_parens && (match ex with ExCtor _ -> true | _ -> false)
+  let needs_wrap = match ex with
+    | ExCtor(_, _ :: _, _) -> true  (* has named params *)
+    | ExCtor(_, _, _ :: _) -> true  (* has positional args *)
+    | _ -> false
+  in
+  if need_parens && needs_wrap
   then "(" ^ pp ^ ")"
   else pp
 
--- a/test/err/pm_0001_nonexhaustive_nested.fram
+++ b/test/err/pm_0001_nonexhaustive_nested.fram
@@ -14,4 +14,4 @@ let area = fn fig =>
   end
 
 # @stderr:not exhaustive
-# @stderr:Many (MyCons (Square) _)
+# @stderr:Many (MyCons Square _)
--- a/test/err/pm_0006_nonexhaustive_infix.fram
+++ b/test/err/pm_0006_nonexhaustive_infix.fram
@@ -1,0 +1,11 @@
+data rec List X = [] | (::) of X, List X 
+data rec List2 X = Nil | Cons of X, List2 X 
+
+let lst = 1 :: 2 :: 3 :: []
+let len = match lst with 
+  | [] => 0
+  | x :: [] => 1 
+end
+
+# @stderr:not exhaustive
+# @stderr:(::) _ ((::) _ _)
--- a/test/err/pm_0006_nonexhaustive_nested.fram
+++ b/test/err/pm_0006_nonexhaustive_nested.fram
@@ -1,0 +1,12 @@
+data D = D1 | D2
+data C = C of D
+data B = B of C
+data A = A of B
+
+let f x =
+  match x with
+  | A (B (C D1)) => 0
+  end
+
+# @stderr:not exhaustive
+# @stderr:A (B (C D2))


src/EffectInference/Error.ml
```diff
@@ -120,9 +120,11 @@ let escaping_effect_var ~origin x =
   (pos, msg1 ^ msg2 ^ T.Pretty.additional_info pp_ctx)
 
 let non_exhaustive_match ~pos ctx =
-  (* TODO: counterexample in the message. *)
+  let counterex = PatternContext.to_pattern ctx in
+  let ex_str = PatternContext.pp_ex_pattern counterex in
   let msg = Printf.sprintf
-    "This pattern-matching is not exhaustive."
+    "This pattern-matching is not exhaustive.\nHere is an example of a case that is not matched:\n  %s"
```
ppolesiuk: This line is longer than 80 columns. I suggest to break it.
```suggestion
    "This pattern-matching is not exhaustive.\n\
    Here is an example of a case that is not matched:\n  %s"
```

src/EffectInference/PatternContext.ml
```diff
@@ -94,3 +94,46 @@ let focus_with ctx ex =
...
+  | CtxDone ex -> ex
+  | _ ->
+    match refocus_with ctx ExWildcard with
+    | CtxDone ex -> ex
+    | _ -> ExWildcard (* Should not happen, but safe fallback *)
+
+(** Pretty-print a name *)
+let pp_name (name : T.name) =
+  match name with
+  | NVar s | NOptionalVar s | NImplicit s | NMethod s -> s
```
ppolesiuk: Methods names are not printed correctly. They should be printed as `method m = ...`. For optional parameters in patterns, `x = p` and `?x = p` means the same, it would be better to stress that the parameter is optional, and use the latter form.
```suggestion
  | NVar s | NImplicit s -> x
  | NOptionalVar s -> "?" ^ s
   | NMethod s -> "method " ^ s
```

src/EffectInference/PatternContext.ml
```diff
@@ -94,3 +94,46 @@ let focus_with ctx ex =
...
+and pp_ex_pattern ex =
+  match ex with
+  | ExHole -> "_"
+  | ExWildcard -> "_"
+  | ExCtor(name, ims, args) ->
+    let pp_named (n, ex) =
+      pp_name n ^ " = " ^ pp_ex_pattern ex
+    in
+    let named_str = String.concat ", " (List.map pp_named ims) in
+    match ims, args with
```
ppolesiuk: I would consider omitting named parameters with trivial patterns. This would reduce clutter in the output message.
```suggestion
    let pp_nontrivial_named (n, ex) =
      match ex with
      | ExHole | ExWildcard -> None
      | ExCtor _ -> Some (pp_name n ^ " = " ^ pp_ex_pattern ex)
    in
    let ims_strs = List.filter_map pp_nontrivial_named ims in
    let named_str = String.concat ", " ims_strs in
    match ims_strs, args with
```

src/EffectInference/PatternContext.ml
```diff
@@ -94,3 +94,46 @@ let focus_with ctx ex =
...
+let pp_name (name : T.name) =
+  match name with
+  | NVar s | NOptionalVar s | NImplicit s | NMethod s -> s
+
+(** Pretty-print an example pattern, wrapping in parens if needed *)
+let rec pp_ex_pattern_wrap need_parens ex =
+  let pp = pp_ex_pattern ex in
+  if need_parens && (match ex with ExCtor _ -> true | _ -> false)
+  then "(" ^ pp ^ ")"
+  else pp
```
ppolesiuk: The implemented logic of adding parentheses scales badly to more than two precedence levels. It would be better to pass the current precedence level to the `pp_ex_pattern` and for each case check if parentheses are needed (using some auxiliary function). See for instance the implementation of `print_type` in `src/Lang/UnifCommon/Pretty.ml`. We would need more precedence levels to correctly print operator-like constructors, e.g., `(::)`.

test/err/pm_0001_nonexhaustive_nested.fram
```diff
@@ -0,0 +1,17 @@
+data Shape = Circle | Square | Triangle
+
+data rec MyList T = Nil | MyCons of T, (MyList T)
```
ppolesiuk: Parentheses are not needed here.
```suggestion
data rec MyList T = Nil | MyCons of T, MyList T
```

rureirureirurei:
@ppolesiuk could you please review the updates?

ppolesiuk (approved):
As I said in one of comments, the logic behind adding parenthesis to patterns scales badly. However, this issue will be important, once we will add support of operator-like constructors in the printer, so it will be fixed later. Other changes look solid, and are ready for the merge.

PR #283, opened 2025-11-24
Disallow higher-order effect kinds

### LINKED ISSUE #199: Disallow higher-order effect kinds.
Type-checker allows arrow kinds with `effect` codomain, but later phases assume that such situation never occurs. This may result in internal errors. For instance, running the program
```
data T E = { run : Unit ->[E Unit] Unit }
```
results in
```
Fatal error: exception Failure("Internal error: type application in effect")
```

Allowing such kinds in later phases seems to be hard task, and I don't think we really need it. Therefore I opt for disallowing such kinds in the type-checking phase. I see the following two possible solutions to this problem (I prefer the second).
- We can either split grammar of kind into two syntactic categories to enforce that `effect` never occurs on the right-hand-side of an arrow;
- or we can reintroduce constraints on kind unification variables, similar to those enforcing that there are no unification variables of `effect` kind in the old type-checker.

ppolesiuk:
Now, the type-checker tracks which kind unification variables cannot be set to effect kind. Such a requirement may appear on right-hand-side of arrow kinds, because the effect inference algorithm requires that all atomic effects are effect variables.

The error messages are not perfect. For instance in the motivating example:
```
data T E = { run : Unit ->[E Unit] Unit }
```
we get the following error:
```
fatal error: This type has kind ?k0, but it was expected of kind effect
   | data T E = { run : Unit ->[E Unit] Unit } ;;
   |                            ^^^^^^
   | 
```

Fixes #199

### DIFF
--- a/src/ToCore/Common.ml
+++ b/src/ToCore/Common.ml
@@ -7,7 +7,8 @@
 module S = Lang.ConE
 module T = Lang.Core
 
-(** Translate kind *)
+(** Translate kind. It also instantiates all kind unification variables to
+  [KType], since they are not representable in the Core language. *)
 let rec tr_kind k =
   match Lang.Unif.Kind.view k with
   | KType   -> T.Kind.Ex KType


src/ToCore/Common.ml
```diff
@@ -13,7 +13,8 @@ let rec tr_kind k =
   | KType   -> T.Kind.Ex KType
   | KEffect -> T.Kind.Ex KEffect
   | KUVar u ->
-    Lang.Unif.KUVar.set u Lang.Unif.Kind.k_type;
+    let ok = Lang.Unif.KUVar.set u Lang.Unif.Kind.k_type in
+    assert ok;
```
Foxinio: This `tr_kind` function translates kind from ConE representation to Core representation, but also sets every unsolved UVar to be of kind type. I know that this is a fix faithful to previous implementation, but should this behaviour be better documented.

PR #287, opened 2025-12-04
Add or-patterns support in pattern matching

### LINKED ISSUE #208: Or patterns
I would be nice to support "or" patterns. Some work towards this direction was done: checking patterns in a type-checker return partial environments, that can be manipulated in some ways.

zielinsky:
Closes https://github.com/fram-lang/dbl/issues/208


src/TypeInference/Error.ml
```diff
@@ -479,6 +479,38 @@ let duplicate_module_in_pattern ~pos ~ppos name =
...
+  in
+  (pos, msg ^ T.Pretty.additional_info pp_ctx, [])
+
+let or_pattern_scheme_mismatch ~pos ~pp sch1 sch2 =
+  let pp_ctx = T.Pretty.empty_context () in
+  let msg = Printf.sprintf
+    ("Or-pattern branches must bind the same variables with the same types."
+    ^^ " A variable has type %s in one branch, but %s in another.")
+    (T.Pretty.pp_scheme pp_ctx pp sch2)
+    (T.Pretty.pp_scheme pp_ctx pp sch1)
```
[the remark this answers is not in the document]
zielinsky: Just a cosmetic change, but okay - done.

src/DblParser/YaccParser.mly
```diff
@@ -465,10 +465,22 @@ str_interp
...
+
+pattern_no_or
+: expr { $1 }
+;
+
 /* ========================================================================= */
 
 match_clause
-: expr ARROW2 expr { make (Clause($1, $3)) }
+: pattern_or ARROW2 expr { make (Clause($1, $3)) }
```
ppolesiuk: So, or-patterns can be used only at the level of match-clauses. Nested or patterns, like `Some(p1 | p2)` are not recognized by the parser. There is no reason for such a restriction.

src/EffectInference/PatternMatch.ml
```diff
@@ -97,8 +97,18 @@ let rec simplify_as_pattern x (cl : iclause) =
...
+  | POr(pat1, pat2) :: pats ->
+    expand_or_at_head { cl with c_patterns = pat1 :: pats } @
+    expand_or_at_head { cl with c_patterns = pat2 :: pats }
+  | _ -> [cl]
+
 (** Simplify as-patterns on the head position in given clause list *)
 let simplify_as_patterns x cls =
+  (* First expand or-patterns, then simplify as-patterns *)
+  let cls = List.concat_map expand_or_at_head cls in
   List.map (simplify_as_pattern x) cls
```
ppolesiuk: I'm afraid that it would not work, when as-patterns are interleaved with or-patterns. In such a case, the algorithm would return non simplified or-pattern, which will result in assertion fail in line 146. However, I cannot test it, because the parser doesn't allow to use or-patterns under as-patterns.

ppolesiuk (changes requested):
I like this direction, but some things should be fixed. The most important ones are:
- parser
- proper handling of existential types in patterns.

ppolesiuk (changes requested):

ppolesiuk (commented):
And of course, merge conflicts should be resolved.

src/TypeInference/Error.ml
ppolesiuk, line 1: Is it possible to make this code fit into 80 columns?

src/TypeInference/PartialEnv.ml
```diff
@@ -212,6 +212,99 @@ let join ~pp penv1 penv2 =
...
+  in
+  StrMap.fold (fun mname minfo2 ren ->
+    match StrMap.find_opt mname penv1.module_map with
+    | Some minfo1 ->
+      let vars1 =
+        List.fold_left
+          (fun m (name, var, _) -> Name.Map.add name var m)
+          Name.Map.empty minfo1.mi_vals
+      in
+      List.fold_left (fun ren (name, var2, _) -> add_var_ren vars1 name var2 ren) ren minfo2.mi_vals
```
ppolesiuk: This line is very long.

src/TypeInference/PartialEnv.ml
```diff
@@ -212,6 +212,99 @@ let join ~pp penv1 penv2 =
...
+     not (T.TVar.Map.is_empty penv2.tvar_tab) then
+    Error.report (Error.or_pattern_binds_type_vars ~pos);
+  let ren = build_ren ~scope penv1 penv2 in
+  let penv =
+    { tvar_map   =
+        StrMap.merge (intersect_tvar_info ~pos) penv1.tvar_map penv2.tvar_map;
+      tvar_tab   =
+        T.TVar.Map.empty;
+      val_map    =
+        Name.Map.merge (intersect_val_info ~check_schemes ~pp ~pos) penv1.val_map penv2.val_map;
```
ppolesiuk: And this...

src/TypeInference/PartialEnv.ml
```diff
@@ -212,6 +212,99 @@ let join ~pp penv1 penv2 =
...
+      module_map =
+        StrMap.merge (intersect_module_info ~check_schemes ~pp ~pos) penv1.module_map penv2.module_map
```
ppolesiuk: And this...

src/DblParser/YaccParser.mly
```diff
@@ -479,10 +479,18 @@ str_interp
 | expr_40             ESTR            {[($1, None,    $2)]      }
 ;
 
+/* ========================================================================= */
+/* PATTERNS */
```
ppolesiuk: While, the implementation is correct, I would consider renaming `pattern` into `pattern_expr` or something similar, to stress that this construct is not only for patterns, but also for expressions (and desugaring phase would report invalid use of `BAR`). I would also move the definition at the beginning of the block with expression rules.
```suggestion
/* Pattern-level expressions, i.e., the level lower then the level of definitions. */
```

ppolesiuk (commented):
Finally, the implementation looks great! I still have some minor comments regarding code formatting, but if these will be addressed, I would merge the PR.

src/TypeInference/Error.ml
zielinsky: done in https://github.com/fram-lang/dbl/pull/287/commits/2cdf6f29cd061a6ad23ff5e847afea91289febf5

src/TypeInference/PartialEnv.ml
zielinsky: done in https://github.com/fram-lang/dbl/pull/287/commits/2cdf6f29cd061a6ad23ff5e847afea91289febf5

src/TypeInference/PartialEnv.ml
zielinsky: done in https://github.com/fram-lang/dbl/pull/287/commits/2cdf6f29cd061a6ad23ff5e847afea91289febf5

src/TypeInference/PartialEnv.ml
zielinsky: done in https://github.com/fram-lang/dbl/pull/287/commits/2cdf6f29cd061a6ad23ff5e847afea91289febf5

src/DblParser/YaccParser.mly
zielinsky: done in https://github.com/fram-lang/dbl/pull/287/commits/220bf8beb05e7dcd8486dc566333ada662712118

PR #289, opened 2025-12-06
Testing framework

wojpok:
A quick draft I wrote. Documentation is missing and will be written soon. There are issues with `~__file__` implicit. Upon nesting, this implicit stops showing any meaningful location, which is a problem with code position tracking. Maybe separate implicit is needed?

### DIFF
--- a/lib/String.fram
+++ b/lib/String.fram
@@ -9,9 +9,10 @@
 
   @asserts Range must be contained within string.
  ##}
-pub let substring (str : String) (pos : Int) (len : Int) =
+pub let substring
+    {~__line__, ~__file__} (str : String) (pos : Int) (len : Int) =
   assert
-    {msg="asdsad"}
+    {msg="Substring range is outside string."}
     (pos >= 0 || len >= 0 || pos + len <= str.length);
   let substr = extern dbl_subStr : String -> Int -> Int ->[] String in
   substr str pos len
--- a/lib/Testing.fram
+++ b/lib/Testing.fram
@@ -53,7 +53,6 @@ section
     end
 
   let extendPath (name : String) =
-    printStrLn ~__modulePath__;
     let pth = getPath () in
     Path {path = Some (pth + [name])}
 end


lib/Base/Option.fram
```diff
@@ -38,3 +38,10 @@ pub method map self f =
   | None   => None
   | Some x => Some (f x)
   end
+
+pub method equal {X, method equal : X -> X ->[] Bool} (o1 : Option X) o2 =
+  match (o1, o2) with
+  | (Some x1, Some x2) => x1.equal x2
```
Foxinio:
```suggestion
  | (Some x1, Some x2) => x1 == x2
```

lib/String.fram
```diff
@@ -0,0 +1,58 @@
...
+    Some (charListToStr (iter (pos + len - 1) [])))
+
+{##
+  Returns a index of first ocurrence of given char and `None` if
+  such position was not found.
+
+  @param ?beg Initial search position. Allows to iteratively
+    find all ocurrences of given char.
+ ##}
+pub let findChar {?beg} (str : String) (chr : Char) =
```
Foxinio: I would consider a different name for the parameter, like full `begin`. I don't think we are saving the world by conserving this amount of space, and I had to read the documentation to understand what this parameter does, and then think a second to understand why is it called that.

lib/Testing.fram
```diff
@@ -0,0 +1,551 @@
...
+
+# Source code location tracker
+data Loc = {file : String, line : Int}
+
+# Position smart constructor
+let mkLoc {~__line__ = line, ~__file__ = file} = Loc {line, file}
+
+method toString (Loc {file, line}) = "\{file}:\{line}"
+
+# ====== Pathing ======
```
Foxinio: Personally I think, separators like this one (and other present throughout this file), have the purpose of helping with quick visual parsing the file structure. To play this role they need to be very noticeable, and these are not.

I propose following the convention present in files like `Map.fram`, so this would look like:
```suggestion
#==============================================================================
# Pathing
#==============================================================================
```

lib/Testing.fram
```diff
@@ -0,0 +1,572 @@
...
+
+## # Testing framework
+
+# =============================================================================
+# Helper functions
+
+# -----------------------------------------------------------------------------
+# Location
+
+# Source code location tracker
```
Foxinio: Here and throughout the file, please use double `#` for documenting comments

lib/String.fram
```diff
@@ -0,0 +1,55 @@
...
+ # See LICENSE for details.
+ #}
+
+## # String operations 
+
+{##
+  Returns a substring of length `len` starting at index `pos`. Returns `None`
+  when requested range is not contained within string.
+ ##}
+pub let substring (str : String) (pos : Int) (len : Int) =
```
ppolesiuk: Should this function be so defensive? Taking a substring is similar to array indexing, so we could similarly abort the program if the requested range is out of bounds. I cannot imagine a scenario, where the user could match on the result, and do something important, when this function returns `None`.

ppolesiuk (approved):
I'm really impressed how beautiful this code is. It uses all nice features of Fram (effect handlers, sections, implicit parameters, methods, ...) to encode the complex logic in very elegant and readable way. I have only two minor remarks regarding proposed extensions to the standard library.

PR #290, opened 2025-12-07
Fixed bug in simplification of subeffecting constraints

ppolesiuk:
Fixed bug in constraint simplification. Additionally, S-expression printers for the internal representation of constraints were implemented, in order to help potential future debugging of the constraint simplification procedure.


forell (approved):
Looks reasonable. At a quick glance, it doesn't seem like any other bits of simplification suffer from the same bug.

PR #301, opened 2025-12-23
Add mssing jsonrpc fields to LSP

dominik-muc:
This was conflicting with LSP specification and thus Zed extension couldn't communicate with framls.


wojpok (approved):
LGTM

PR #307, opened 2026-01-15
Mut handler in test framework

wojpok:
This PR introduces pure mutability effect to the test case environment. The Mutability test file have been adjusted as a proof of concept. Method tests could be improved, but it is a good middle ground in test refactor


kulson (approved):
I approve this pull request. It helped me test the vector module, and everything works as expected.

PR #319, opened 2026-01-31
Effect annotations

### LINKED ISSUE #127: Effect annotations
Expressions can be annotated with types (`e : τ`). It would be useful to annotate them with effects. We have to agree on concrete syntax. I see the two options (I prefer the first one):

1. In order to be consistent with syntax of arrows, we can have `e : [Eff] Typ` syntax.
2. We can follow the notations from formal metatheory and syntax of Helium: `e : Typ / Eff`.

In both cases there is a question how to annotate that the expression is strongly pure (it always terminates). Maybe just by omitting the effect?

ppolesiuk:
This PR introduces effect annotations in the grammar of expressions and result type/effect annotations in let and method definitions.

Fixes #127

### DIFF
--- a/src/DblParser/Desugar.ml
+++ b/src/DblParser/Desugar.ml
@@ -51,8 +51,13 @@ let node_is_data_def (def : def) =
 
 let annot_tp e tp =
   { pos = e.pos;
-    data = Raw.EAnnot(e, with_nowhere tp)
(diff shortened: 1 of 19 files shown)


forell (approved):
Looks good, I appreciate the change to "total" in user-facing errors.

PR #320, opened 2026-02-03
Updated Prelude tests

wojpok:


wojpok:
I double checked everything - there are no issues with closing parenthesis. Moreover - effect systems won't allow me to define a test case within a test case

ppolesiuk (approved):
Looks good. Thanks!

PR #322, opened 2026-02-09
Updated Map tests

wojpok:


wojpok:
The randomized test is really resource heavy. I had to increase the timeout to fit this in. Since we are going to replace this script soon, I think this temporary solution is sufficient

Issue #316, opened 2026-01-30
Homogeneous types of arithmetic operators


ppolesiuk:
Arithmetic operators are defined as polymorphic functions, that expects that the type parameter has corresponding method. For instance, the addition is defined as follows:
```fram
pub let (+) {type T, method add : T -> _} (x : T) = x.add
```
This implementation allows totality of addition, but requires additional type annotations to use it:
```fram
let f x = x + 1
```
ends up with the following error:
```fram
fatal error: Cannot resolve a method add associated with type #?T
...
```

An alternative approach would be to enforce that the arithmetic operators have homogeneous type:
```
pub let (+) {type T, method add : T -> T ->> _} (x : T) = x.add
```
This solution has two drawbacks: the `(+)` function is not total, and its type is less polymorphic. However these drawbacks should not pose problems in practice, as (1). totality is an internal tool for ensuring productivity and approximating purity restriction, and is not intended to be overused by a programmer; (2). real arithmetic operators are homogeneous. On the other hand, homogeneous operators makes type inference works better. For instance, functions like
```
let f x = x + 1
```
would be accepted by the type-checker.

During the last meeting we agreed that we prefer to have homogeneous arithmetic operators in the standard library.

PR #323, opened 2026-02-09
Function combinators

wojpok:


lib/Prelude.fram
```diff
@@ -73,3 +73,6 @@ pub let rec replicate (s : String) (n : Int) =
   else
    (let s' = replicate s (n / 2) in
     s' + (s' + s))
+
+## Function composition method.
+pub method mul {X, Y, Z} (f : Y ->> Z) (g : X ->> Y) (x : X) = f (g x)
```
ppolesiuk: As we decided on the meeting, the arithmetic-like operators should have homogeneous types (see #316). Therefore, we should not reuse multiplication as a function composition, because it would be too restrictive. We need to find another operator. We could ask people on discord, which one is the best choice.

PR #335, opened 2026-03-25
First class handlers from handling functions

### LINKED ISSUE #331: First class handlers from handing functions
It would be nice to allow defining first class handlers by providing handling function. We agreed (more-less) on the following syntax:
```
handlerfn
  <def₁> ... <defₙ>
  in
  <expr_cap>
  return <pat_r> => <expr_r>
  finally <pat_f> => <expr_f>
end
```
The return and finally clauses are optional. The meaning of such a code is a first class handler that realizes the following handling function:
```
fn run =>
  let <pat_f> =
    <def₁> ... <defₙ>
    let <pat_r> = run <expr_cap>
    in
    <expr_r>
  in
  <expr_f>
```

ppolesiuk:
This PR provides a new way of defining first class handlers on top of a sequence of definitions and an effect capability. With this construct it is possible to map capability, input and output types of an existing first-class handler. For instance, it is possible to define a new first-class handler on top of the existing one, by wrapping the capability around a constructor of some, possibly abstract datatype.

Proposed implementation doesn't support explicit instantiation of the handled effect, as it was proposed during the meeting (yes, I know, I didn't write the proposed syntax in the issue yet).

Resolves #331

### DIFF
--- a/src/DblParser/Desugar.ml
+++ b/src/DblParser/Desugar.ml
@@ -292,9 +292,10 @@ let rec tr_ctor_pattern (p : Raw.expr) =
   | ESelect(path, p) -> path_append path (tr_ctor_pattern p)
 
   | EWildcard | ENum _ | ENum64 _ | EStr _ | EChr _ | EParen _ | EVar _
-  | EImplicit _ | EFn _ | EApp _ | EDefs _ | EMatch _ | EHandler _ | EEffect _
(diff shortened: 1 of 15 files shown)


forell (approved):
The `handler` construct should probably not check for totality anymore to be consistent with `handlerfn`, but that can be a separate PR. Keeping both constructs in the intermediate languages for as long as they are feels a bit redundant, but one can't be expressed in terms of the other in `Unif` yet, so that can also wait for when reset0 is available.

PR #336, opened 2026-04-01
Affine effect inference

ppolesiuk:
This PR introduces affine effects to the effect system and adapts the effect inference algorithm to support this feature. However, the algorithm doesn't check if the `aff effect` construct uses the continuation in the affine way–it just trust the programmer.

The implementation is surprisingly simple: it is enough to change the propositional logic used by the effect inference. It is so simple, that I am not totally convinced if it is correct: the simplification rules that manipulates formulas looks reasonable for me, but it would be nice if someone else would look at them with some dose of criticism.

To have affine `IO` effect, effect variables (in Unif and ConE) have attached effect mode, and each use of this variable is implicitly projected to that mode (and this projection is not printed by the pretty-printer). This approach enables simple implementation, and paves the path towards a nice feature: the programmer could annotate type variables with a mode in binders, so the variable is used only with that (or more restrictive) mode. We can discuss this direction later.

As the next steps, the following things should be done (as separate issues):
- the standard library (especially the `Mutable` module) should be adapted to use this feature to improve the interface;
- affine shift₀ should actually check, if the continuation is used in the affine way;
- extend the Core language to support affinity in some way;
- provide a affinity-specific constraint simplification rules;
- extend the system by another effect modes (linear, exceptions).


src/EffectInference/ConstrSimplify.ml
```diff
@@ -112,7 +112,7 @@ let add_irrelevant_constr st c =
   that normalizes the constraints again. *)
 let get_gvar eff =
   match T.Effct.view eff with
-  | ([], [(gv, p)]) when IncrSAT.Formula.is_true p -> gv
+  | ([], [(gv, p)]) when T.Formula.is_true p -> gv
```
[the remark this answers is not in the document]
ppolesiuk: WHAT!? This function shall be used only on effects constructed by `T.Effct.gvar gv`, so it is guaranteed that the formula is true (unless, the variable was set in a meantime). This is unrelated to effect modes.

test/ok/ok0158_affineIdEffect.fram
```diff
@@ -0,0 +1,4 @@
+data Id E = { id : {X} -> X -> [aff E] X }
```
Foxinio: Formatting 
```suggestion
data Id E = { id : {X} -> X ->[aff E] X }
```

test/ok/ok0163_mixedAffinity.fram
```diff
@@ -0,0 +1,11 @@
+data BT E =
+  { flip : Unit ->[E] Bool
+  , fail : {X} -> Unit ->[aff E] X
+  }
+
+let hBT =
+  handler BT
+    {     effect flip () = resume True; resume False
```
Foxinio, line 8: I don't know it it's a good practice, but I maybe we should add a TODO comment, to copy this file with flip incorrectly labeled as aff to error tests once this detection is implemented.

src/EffectInference/Expr.ml
```diff
@@ -495,7 +495,8 @@ and infer_type : type ed.
         () in
     (res, tp, return_pure eff_req)
 
-  | EEffect(lbl, x, body, res_tp) ->
+  | EEffect(lbl, mode, x, body, res_tp) ->
+    (* TODO: check if the usage of resumption respects the mode *)
```
Foxinio, line 529: In spite of what are current guidelines, I suggest including nr of issue in comments such as this one.

src/Lang/ConEPriv/Effct.mli
```diff
@@ -58,15 +55,19 @@ val var : TVar.t -> t
...
 
 (** Create effect that is equal to the given effect when the formula is
   satisfied, and equal to the pure effect when the formula is not satisfied.
   *)
-val guard : t -> formula -> t
+val guard : t -> Formula.t -> t
+
+(** Project the effect to the given mode. The call to [proj mode eff] is
+  equivalent to [guard eff (Formula.of_mode mode)]. *)
+val proj : EffectMode.t -> t -> t
```
Foxinio, line 70: From the first part of the comment i got the impression that this function performs some kind of pass through the effect and sets all leaves (or formulas inside) to have mode given to this function (as one would do in imperative language). Instead this is achieved lazily by adding node that when read is to be interpreted, as if pass through was performed (as is suggested by  the second part of this comment)..
I have no suggestion in this comment, just wanted to confirm i have correct understanding of the code.

Foxinio (commented):
For the first look at this PR, the changes look good.
I did leave some minor suggestions, which relate to code presentation.

However I did not look thoroughly enough at implementation of constraint simplification (which i intend to do shortly).

test/ok/ok0163_mixedAffinity.fram
ppolesiuk: I think it is a bad practice. Nobody would look into test files as long as they work correctly, so such TODO comment would stay there forever. On the other hand, the person who will implement this detection (probably me) would be responsible to provide tests, and probably would write such a test, or very similar one from the scratch, as it is very obvious thing to check.

src/EffectInference/Expr.ml
ppolesiuk: Are there any such guidelines? I see a cyclic dependency: to get the issue number we need to create the issue, for which we need to merge this PR, for which we should update this comment, but this require the issue number.

src/Lang/ConEPriv/Effct.mli
ppolesiuk: Since the function takes the effect and returns the effect, it at least pretends to be purely functional: it doesn't alter the meaning of the input effect. However, propositional variables, or gvars might be set during the effect inference process, thus the exact representation of the effect may change each time it is viewed, by removing some redundant parts. It is just an optimization, in the same spirit as path compression in union-find, or in the type representation in the Unif language.

forell (approved):
I'm happy with this implementation and in favor of merging it before the actual checking of `aff effect` is implemented.

Issue #286, opened 2025-11-28
Concrete syntax for handler types


ppolesiuk:
We should extend the parser (and fix the type pretty-printer) to support handler types. We agreed on the following syntax:
```
handler {E} T_cap in [E_in] T_in => [E_out] T_out
```
Optionally, the binder `{E}` and effects can be omitted. Moreover, it would be nice to support the shortened form
```
handler {E} T_cap => [E_out] T_out
```
when `[E_in] T_in` and `[E_out] T_out` are (pairwise) the same.

PR #337, opened 2026-04-12
Concrete syntax for handler types

JakubZojdzik:
#286


src/TypeInference/Type.ml
```diff
@@ -50,7 +50,12 @@ let rec infer_kind env (tp : S.type_expr) =
 
   | THandler { effct; cap_type; in_type; in_eff; out_type; out_eff } ->
     let (in_env, _) = Env.enter_scope env in
-    let (in_env, eff_var) = Env.add_tvar ~pos in_env effct T.Kind.k_effect in
+    let (in_env, eff_var) =
+      if String.equal effct "_" then
```
JakubZojdzik: Couldn't think of anything more elegant as `effct` is expected to be string everywhere

ppolesiuk: Generally, it is good practice to avoid if-expressions, where the problem can be solved by pattern-matching. It would be better to change the definition of `THandler` constructor and have `effct` field of the `option` type.

src/Lang/UnifCommon/Pretty.ml
```diff
@@ -354,25 +354,38 @@ let rec print_type env (tp : type_tree) prec =
 
   | PP_THandler h ->
     paren env prec 0 (fun () ->
-      let name = gen_tvar_name (fun n -> fresh_for_type env n tp) "E" in
+      let name =
+        match Env.lookup_tvar env h.eff_var with
+        | Found name -> name
+        | Anon(name, _) -> name
+        | Unbound _ -> gen_tvar_name (fun n -> fresh_for_type env n tp) "E"
+      in
```
ppolesiuk: There is no point in looking up the environment here, as the variable is not bound in this scope (it would be bound in the next line). I know, that asking the environment for a name that is meaningful for the programmer is tempting, but the binding structure doesn't work that way. Such an programmer-friendly name could be stored in a variable metadata, but we could do this a separate issue.
```suggestion
      let name = gen_tvar_name (fun n -> fresh_for_type env n tp) "E" in
```

src/Lang/UnifCommon/Pretty.ml
```diff
@@ -354,25 +354,38 @@ let rec print_type env (tp : type_tree) prec =
...
+        | [ PP_EffWildcard ] -> ()
+        | _ ->
+          Env.print_string env " [";
+          print_effect_body env eff;
+          Env.print_string env "]"
+        end;
+        Env.print_string env " ";
+        print_type env tp 1
+      in
+      if h.in_eff = h.out_eff && h.in_tp = h.out_tp then (
```
ppolesiuk: To be fully correct, we should also check, that `h.eff_var` doesn't appear in inner type and effect, because `handler {E} T in [E] _ => [E] _` doesn't mean the same as `handler {E} T => [E] _`. However, such a case is extremely rare, and I'm not able to write an example in the concrete syntax that would exhibit this error.

ppolesiuk (commented):
This PR looks good. I have only two minor comments regarding pretty-printing.

src/DblParser/YaccParser.mly
```diff
@@ -168,8 +168,19 @@ ty_expr_simple
...
+| KW_HANDLER effct_bind ty_expr KW_IN ty_expr ARROW2 ty_expr
+  { make (THandler ($2, $3, $5, $7)) }
+| KW_HANDLER effct_bind ty_expr ARROW2 ty_expr
+  { make (THandler ($2, $3, $5, $5)) }
 ;
 
+effct_bind
+: CBR_OPN UID CBR_CLS { Some $2 }
+| CBR_OPN UNDERSCORE CBR_CLS { None }
+| /* empty */         { None }
```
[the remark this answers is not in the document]
ppolesiuk: I agree. Proposed change should remove shift-reduce conflicts, or at least reduce their number.

src/DblParser/YaccParser.mly
```diff
@@ -168,8 +168,19 @@ ty_expr_simple
 | UNDERSCORE { make TWildcard }
 | SBR_OPN effct SBR_CLS { make ($2).data }
 | CBR_OPN ty_field_list CBR_CLS { make (TRecord $2) }
+| KW_HANDLER effct_bind ty_expr KW_IN ty_expr ARROW2 ty_expr
+  { make (THandler ($2, $3, $5, $7)) }
+| KW_HANDLER effct_bind ty_expr ARROW2 ty_expr
+  { make (THandler ($2, $3, $5, $5)) }
 ;
```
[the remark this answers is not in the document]
ppolesiuk: I agree. It would be nice to have some tests for the new feature.

src/Lang/UnifCommon/Pretty.ml
ppolesiuk: Moreover, structural equality is bad (for many reasons). Maybe a custom function that checks α-equivalence modulo permutations of effects, and understands, that the scope of compared types/effects are different would solve all the problems.

JakubZojdzik:
I had some troubles on implementing α-equivalence so I ended up checking if bounded effect is used in the output type and the structural equivalence of both types. Also, current solution looks over complicated and is not scalable easily - every update in type_expr affects the sorting order so as the effects comparison. If needed, I can spend more time implementing a proper α-equivalence test, but if it is "good enough" I would consider keeping pretty printer simple and always printing the expanded type.

src/DblParser/YaccParser.mly
```diff
@@ -174,8 +174,22 @@ ty_expr_simple
...
+| KW_HANDLER ty_expr ARROW2 ty_expr
+  { make (THandler (None, $2, $4, $4)) }
 ;
 
+effct_bind
+: CBR_OPN UID CBR_CLS { Some $2 }
+| CBR_OPN UNDERSCORE CBR_CLS { None }
+;
+
+
```
ppolesiuk: Double empty line

src/DblParser/Desugar.ml
```diff
@@ -138,6 +138,24 @@ let rec tr_type_expr (tp : Raw.type_expr) =
...
+        Option.value out_eff_opt ~default:(make (TEffect [ make TWildcard ]))
+      in
+      make (THandler {
+        effct = effct_opt;
+        cap_type;
+        in_type;
+        in_eff;
+        out_type;
+        out_eff;
+      })
```
ppolesiuk: Please keep code formatting consistent. Also, the record at the end of the clause could be a bit more concise.

```suggestion
    let cap_type = tr_type_expr cap_tp in
    let in_eff_opt, in_type = tr_eff_type in_comp in
    let in_eff =
      Option.value in_eff_opt ~default:(make (TEffect [ make TWildcard ]))
    in
    let out_eff_opt, out_type = tr_eff_type out_comp in
    let out_eff =
      Option.value out_eff_opt ~default:(make (TEffect [ make TWildcard ]))
    in
    make (THandler {
      effct = effct_opt; cap_type;
      in_type;  in_eff; out_type; out_eff;
    })
```

src/Lang/UnifCommon/Names.mli
```diff
@@ -0,0 +1,23 @@
...
+  | NVar of string
+  | NOptionalVar of string
+  | NImplicit of string
+  | NMethod of string
+
+val compare_tname : tname -> tname -> int
+
+val compare : name -> name -> int
+
+val equal : name -> name -> bool
```
ppolesiuk, line 26: Function signatures in `*.mli` files should have some documentation comment, even, if it is obvious what they are doing.

src/Lang/UnifCommon/Pretty.ml
```diff
@@ -328,65 +390,75 @@ let print_effect env effs prec =
   match effs with
   | [eff] -> print_single_effect env eff prec
   | _ ->
-    Env.print_string env "[";
-    print_effect_body env effs;
-    Env.print_string env "]"
+      Env.print_string env "[";
+      print_effect_body env effs;
+      Env.print_string env "]"
```
ppolesiuk: Please do not change the formatting of the existing code, unless it makes it more consistent. Same applies to other changes to the end of this file.

src/Utils/PPTree.mli
```diff
@@ -54,4 +54,4 @@ val leave_module : public:bool -> t -> string -> t * pp_module
 val open_module : public:bool -> t -> pp_module -> t
 
 (** Lookup for a unique name of the type of given UID. *)
-val lookup : t -> uid -> pp_result
+val lookup : t -> uid -> pp_result
```
ppolesiuk: No new line at the end of the file.

test/err/tc_0024_handlerMismatch.fram
```diff
@@ -0,0 +1 @@
+let bad {A} {E} (x : handler A => A) : handler {E} A in A => (A ->[E] A) = x
```
ppolesiuk: No newline at the end of the file.

test/err/tc_0025_handlerBinderInOutput.fram
```diff
@@ -0,0 +1 @@
+let bad {A} {E} (x : handler A => A) : handler {E} (A ->[E] A) in (A ->[E] A) => A = x
```
ppolesiuk: No newline at the end of the file.

test/err/tc_0026_handlerBinderInInput.fram
```diff
@@ -0,0 +1 @@
+let bad {A, B, E} (x : handler A => A) : handler {E} B in [E] B => [E] B = x
```
ppolesiuk: No newline at the end of the file.

src/Lang/UnifCommon/Pretty.ml
```diff
@@ -328,65 +390,75 @@ let print_effect env effs prec =
...
+            | _ ->
+                Env.print_string env " [";
+                print_effect_body env eff;
+                Env.print_string env "]"
+            end;
+            Env.print_string env " ";
+            print_type env tp 1
+          in
+          if
+            effects_equal h.in_eff h.out_eff && h.in_tp = h.out_tp
```
ppolesiuk: Should we compare type for equality too? Maybe it would be good to implement structural equality of types (not necessarily modulo α-equivalence), but which uses the order on effects. It would be good placeholder for a better α-equality test.

src/Lang/UnifCommon/Pretty.ml
```diff
@@ -315,6 +315,68 @@ let rec print_single_effect env (eff : effect_tree) prec =
...
+    (match (m1, m2) with
+      | Affine, Affine | Unrestricted, Unrestricted -> compare_effect e1 e2
+      | Unrestricted, _ -> -1
+      | _, Unrestricted -> 1)
+
+let effects_equal effs1 effs2 =
+  if List.length effs1 <> List.length effs2 then false
+  else
+    let sorted1 = List.sort compare_effect effs1 in
+    let sorted2 = List.sort compare_effect effs2 in
```
ppolesiuk: The first test is not needed. Types are not very big, so fighting for effectiveness is pointless. With removed this test, maybe it make sense to use `sort_uniq` instead of `sort`?

ppolesiuk (changes requested):
The code formatting should be improved. I still have some remarks about the α-equivalence.

src/Lang/UnifCommon/Pretty.ml
```diff
@@ -315,6 +315,66 @@ let rec print_single_effect env (eff : effect_tree) prec =
...
+      || tvar_in_type x h.out_tp
+      || tvar_in_effects x h.out_eff
+  | PP_TEffect eff -> tvar_in_effects x eff
+  | PP_TApp (tp1, tp2) -> tvar_in_type x tp1 || tvar_in_type x tp2
+  | PP_TAlias (_, tp) -> tvar_in_type x tp
+
+and tvar_in_scheme x (sch : scheme_tree) =
+  List.exists (fun (_, y) -> TVar.equal x y) sch.ppsch_targs
+  || List.exists (fun (_, s) -> tvar_in_scheme x s) sch.ppsch_named
+  || tvar_in_type x sch.ppsch_body
```
ppolesiuk: `y` is a binding occurrence here, so if `y` is equal to `x` then `x`doesn't have free occurrence in the scheme.
```suggestion
  List.for_all (fun (_, y) -> not (TVar.equal x y)) sch.ppsch_targs &&
  (List.exists (fun (_, s) -> tvar_in_scheme x s) sch.ppsch_named
  || tvar_in_type x sch.ppsch_body)
```

ppolesiuk (approved):
I have one minor comment regarding variable binding structure. The rest looks good. IMO the code is ready to merge, and all remaining issues regarding variable binding (free variables, α-equivalence) can be done as a separate issue. Thanks!

src/DblParser/YaccParser.mly
ppolesiuk, line 1: I cannot merge this PR yet. There are 9 shift/reduce conflicts in the grammar. Please fix them.

ppolesiuk (changes requested):

PR #338, opened 2026-05-01
Basic support for file IO operations

fatmage:
Implemented basic operations for file IO - reading, writing, seeking etc.
Currently unsure about the interface for interacting with result values, happy to discuss it during the meeting.
Closes #334


wojpok (approved):
I love the idea to tie effects with access rights to the file. Interface looks solid and tests are thorough.

There is one minor issue though. Pay attention to the documentation. It is mostly fine, but there are some public and abstract definitions without docstrings. Also remember to end all sentences with a dot and use Present Simple when possible.

Not a huge blocker, I will just approve this right away

PR #342, opened 2026-05-04
Handle non-total handlers

forell:
This makes the usual first-class handlers consistent with handlers from handling functions in #335 by not requiring totality for the effect capability.


ppolesiuk (approved):
LGTM

Issue #344, opened 2026-05-11
Use affine effects in the standard library


forell:
Now that the effect system supports affine effects, it would be nice to update some standard library interfaces (e.g., `Mutable`) to actually use this feature.

PR #351, opened 2026-05-24
Support for affine effects in stdlib

wojpok:
PR adds annotations for affine effects in critical places like Mutable lib and standard effects. This in turn allows to eliminate mutability handler from test framework.

### DIFF
--- a/lib/Mutable.fram
+++ b/lib/Mutable.fram
@@ -76,8 +76,8 @@ pub let ioMut = Mut {E=IO}
 {## Local handler of state.
 
   The mutability effect can be handled locally, provided that the handled
-  computation has no other effects. ##}
-pub let hMut (f : {E : effect} -> Mut E ->[E] _) =
+  computation's effects are affine. ##}
+pub let hMut {type R} (f : {E : effect} -> Mut E ->[E, aff R] _) =
   handle _ / E = () in
   f {E} Mut
 
@@ -86,7 +86,7 @@ pub let hMut (f : {E : effect} -> Mut E ->[E] _) =
   This function is similar to `hMut`, but it allows the handled expression to
   have other effects. However, this function is not pure: it has the `IO`
   effect. ##}
-pub let hMutIO {type R} (f : {E : effect} -> Mut E ->[E,R] _) =
+pub let hMutIO {type R} (f : {E : effect} -> Mut E ->[aff E,R] _) =
   f {E=IO} ioMut
 
 {# ========================================================================= #}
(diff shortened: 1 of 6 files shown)


lib/Mutable.fram
```diff
@@ -76,8 +76,8 @@ pub let ioMut = Mut {E=IO}
 {## Local handler of state.
 
   The mutability effect can be handled locally, provided that the handled
-  computation has no other effects. ##}
-pub let hMut (f : {E : effect} -> Mut E ->[E] _) =
+  computation's effects are affine. ##}
+pub let hMut {type R} (f : {E : effect} -> Mut E ->[aff E, aff R] _) =
```
forell: Any particular reason the new effect parameter is anonymous? I don't feel particularly strongly about this, but I can imagine a situation where I'm not sure why the projection to `aff` fails, so I type out the expected effect in an attempt to debug what's going on.

lib/Mutable.fram
```diff
@@ -86,7 +86,7 @@ pub let hMut (f : {E : effect} -> Mut E ->[E] _) =
   This function is similar to `hMut`, but it allows the handled expression to
   have other effects. However, this function is not pure: it has the `IO`
   effect. ##}
-pub let hMutIO {type R} (f : {E : effect} -> Mut E ->[E,R] _) =
+pub let hMutIO {type R} (f : {E : effect} -> Mut E ->[aff E, aff R] _) =
```
forell: The doc comment for this function is now a bit outdated in its comparison to `hMut`, and with this signature, the whole function is redundant. I'd either remove it altogether or remove the affine projection on `R` to restore the original functionality. I'm not sure which option I prefer.

lib/State.fram
```diff
@@ -30,9 +30,10 @@ pub method update (cell : State E X) f =
...
   handler State
-    { effect get () = fn st => resume st st
-    , effect set st = fn _  => resume () st
+    { aff effect get () = fn st => resume st st
+    , aff effect set st = fn _  => resume () st
     }
   return  x => fn _ => x
   finally c => c initial
   end
+
```
forell: Spurious newline.

lib/Testing.fram
```diff
@@ -130,17 +130,14 @@ let mkTestCase
     (name : String)
     (f : { E_TL : effect
          , E_TT : effect
-         , E_M  : effect
-         , ~mut : M.Mut E_M
-         , ~testLogger : TestLogger E_TL
-         , ~testCaseTracker : TestCaseTracker E_TT
-         } -> Unit ->[E_TL, E_TT, E_M] Unit) =
+         , ~testLogger : TestLogger (aff E_TL)
+         , ~testCaseTracker : TestCaseTracker (aff E_TT)
```
forell: I personally wouldn't use `aff` projections in these parameters, though we don't exactly have any conventions set in stone yet.

forell (changes requested):
Looks good overall, thanks! I left some fairly minor comments.

lib/Mutable.fram
```diff
@@ -208,7 +208,7 @@ pub method toArray {E, ~mut : Mut E} (self : List _) =
 ##}
 pub method pureInitArray
```
ppolesiuk, line 209: For this, and many other functions with `pure` in their name we could give a more general type that allow for any affine ambient effect. Of course, the name should be changed. What do you think?

forell: It'd be nice to either have them updated or at least this mentioned as a comment under issue #344. I guess the issue can stick around until the remaining modes are implemented, for more specific annotations.

PR #352, opened 2026-05-29
Lines exceeding 80 characters

fatmage:
Reformatted lines which were exceeding 80 characters of length.

List of remaining files which break the 80 character rule:
```src/Lsp/Message.ml, line: 19, length: 99
src/Lsp/State.mli, line: 17, length: 120
test/err/tc_0022_or_pattern_mismatch.fram, line: 7, length: 110
test/err/tc_0023_or_pattern_var_mismatch_2.fram, line: 7, length: 110
test/err/tc_0023_or_pattern_var_mismatch_2.fram, line: 8, length: 110
test/err/tc_0024_or_pattern_scheme_mismatch.fram, line: 6, length: 150
test/err/tc_0025_or_pattern_module_mismatch.fram, line: 7, length: 99
```
``Message.ml`` and ``State.mli`` contain a link which is in itself longer than 80 characters, and the test files contain an error message which needs to be validated which is longer than 80 characters.


lib/Testing.fram
```diff
@@ -437,11 +437,12 @@ section
       , Y : type
       , method show : X ->> String
       , method show : Y ->> String
-      , method equal : X ->> Y ->> Bool } 
+      , method equal : X ->> Y ->> Bool }
       (expected : X) (actual : Y) =
     assertTrue
       {msg =
-        "Assertion Failed: Expected \{expected.show} and \{actual.show} to be equal"}
+        "Assertion Failed: Expected \{expected.show}" +
```
wojpok, line 444: Since this line uses string interpolation it can be broken down without string concatenation as such:
```
"Assertion Failed: Expected \{
    expected.show
} and \{
    actual.show
} to be equal"}
```
or
```
"Assertion Failed: Expected \{expected.show
} and \{actual.show} to be equal"}
```
I dont know if and which version is better honestly

lib/Testing.fram
crqch, line 369: Found a typo: `evaulation`

Should be: `evaluation`

I think it can be fixed in this PR since it's about fixing formatting.

PR #353, opened 2026-06-09
Pattern matching for literals

### LINKED ISSUE #137: Pattern matching with exact values.
During the coding of the Hood-Melville queue, I encountered an error message when I had put in a pattern matching an exact value of one of the parameters. 
Simplifying the issue:
1) For this interpreter gives error:
```
data NoneNegativeInt = NNInt of Int

let subOne value = 
match value with
| NNInt 0 => NNInt 0
| NNInt n => NNInt (n-1)
end
```
2. For this interpreter doesn't give error:
```
data NoneNegativeInt = Zero | NNInt of Int

let subOne value = 
match value with
| Zero => Zero
| NNInt n => if n == 1 then Zero else NNInt (n - 1)
end
```
I think it would be great to be able to write matches like in the first point.

lidiapodoluk:
Closes #137, now with requested changes added

### DIFF
--- a/src/Lang/ConE.mli
+++ b/src/Lang/ConE.mli
@@ -127,18 +127,6 @@ type expr =
   | EOptionPrf
     (** ADT-shape proof for option type *)
 
-  | ENum of int
-    (** Integer literal *)
-
-  | ENum64 of int64
-    (** 64 bit integer literal *)
-
-  | EStr of string
-    (** String literal *)
-
-  | EChr of char
-    (** Character literal *)
-
   | ELit of literal
     (** Literal *)
 
@@ -242,7 +230,7 @@ and match_clause =
     (** Body of the clause *)
   }
 
-(** Literals for patterns *)
+(** Literals *)
 and literal =
   | ENum        of int
   | ENum64      of int64
--- a/src/Lang/ConE.mli
+++ b/src/Lang/ConE.mli
@@ -139,6 +139,9 @@ type expr =
   | EChr of char
     (** Character literal *)
 
+  | ELit of literal
+    (** Literal *)
+
   | EVar of var
     (** Variable *)
 
@@ -238,6 +241,13 @@ and match_clause =
     cl_body  : expr
     (** Body of the clause *)
   }
+
+(** Literals for patterns *)
+and literal =
+  | ENum        of int
+  | ENum64      of int64
+  | EStr        of string
+  | EChr        of char
 
 (** Programs *)
 type program = expr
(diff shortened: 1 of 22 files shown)


src/TypeInference/Expr.ml
```diff
@@ -37,33 +37,33 @@ let infer_expr_type ~tcfix ?app_type env (e : S.expr) =
...
-      er_constr = []
-    }
-
-  | EChr c ->
-    { er_expr   = make (T.EChr c);
-      er_type   = Infered (T.Type.t_var T.BuiltinType.tv_char);
-      er_effect = Pure;
-      er_constr = []
-    }
+  | ELit l ->
```
Xaro8, line 40: Unneeded space, after arrow :)

src/TypeInference/ParamResolve.ml
```diff
@@ -244,9 +244,9 @@ and resolve_implicit ~resolve_env rctx iname sch =
     (* Special implicits *)
     let (param_expr, param_tvar) = match iname with
       | "~__line__" -> 
-        (make (T.ENum pos.pos_start_line), T.BuiltinType.tv_int)
+        (make (T.ELit(ENum pos.pos_start_line)), T.BuiltinType.tv_int)
```
Xaro8, line 247: I am not sure, but copilot may be onto something with T.Enum

src/TypeInference/ParamResolve.ml
```diff
@@ -244,9 +244,9 @@ and resolve_implicit ~resolve_env rctx iname sch =
...
       | "~__file__" -> 
-        (make (T.EStr pos.pos_fname), T.BuiltinType.tv_string)
+        (make (T.ELit(EStr pos.pos_fname)), T.BuiltinType.tv_string)
```
Xaro8, line 249: Same as above

src/Lang/ConE.mli
```diff
@@ -116,17 +116,8 @@ type data_def =
...
-    (** ADT-shape proof for unit type *)
-
-  | EBoolPrf
-    (** ADT-shape proof for bool type *)
-
-  | EOptionPrf
-    (** ADT-shape proof for option type *)
-
+(** Literals *)
+and literal =
```
Xaro8, line 120: I suppose and isn't needed there?

test/ok/ok0177_literalMatch.fram
Xaro8, line 1: Maybe a good idea would be to output expectations, for tasting framework

